PR #11, opened 2022-03-10
Feature/auth0 login

LewisDaleUK:


config/urls.py
```diff
@@ -33,8 +33,10 @@
         name="no_results",
     ),
     # Django Admin, use {% url 'admin:index' %}
+    path("accounts/", include("allauth.urls")),
     path(settings.ADMIN_URL, admin.site.urls),
     # Your stuff: custom urls includes go here
+    path("accounts/", include("allauth.urls")),
```
lozette, line 39: Does this need to be in here twice?

LewisDaleUK: Nope 😅

lozette (approved):

PR #16, opened 2022-03-17
Add content flags for sensitive and supplemental 

LornaHarwood:
Added content flags for:

- Case contains sensitive content
- Supplemental documents available

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -22,4 +22,4 @@ requests-toolbelt~=0.9.1
 lxml~=4.7.1
 django-xml~=3.0.0
 wsgi-basic-auth~=1.1.0
-ds-caselaw-marklogic-api-client
+ds-caselaw-marklogic-api-client~=2.0.0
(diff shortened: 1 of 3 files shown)


LewisDaleUK (approved):

requirements/base.txt
```diff
@@ -22,4 +22,4 @@ requests-toolbelt~=0.9.1
 lxml~=4.7.1
 django-xml~=3.0.0
 wsgi-basic-auth~=1.1.0
-ds-caselaw-marklogic-api-client
+ds-caselaw-marklogic-api-client~=2.0.0
```
lozette, line 25: I think this version pin can be removed now

lozette (changes requested):

requirements/base.txt
lozette, line 25:
```suggestion
ds-caselaw-marklogic-api-client
```

lozette (approved):

PR #20, opened 2022-03-18
Amend README to tie this application with ds-caselaw-public-ui

lozette:
This application is intended to work with the Marklogic database from
ds-caselaw-public-ui. However the README instructions gave the impression that
a separate ML instance should be spun up for this app.

I think it's safe to assume that anyone running this application has already
built ds-caselaw-public-ui, so amend the README to make it explicit that the
ML instance this application is the same one as for ds-caselaw-public-ui

Also, remove the ML container from docker-compose, as running this container
conflicts with the one in ds-caselaw-public-ui


lozette:
@LewisDaleUK @Floppy What are your thoughts on this? Is it relatively safe to assume that anyone running this app will also have set up ds-caselaw-public-ui?

LewisDaleUK:
I wonder if it's worth retaining the original instructions, but having a sentence at the beginning (maybe emphasised so that people who are just scanning the README see it) explaining that you don't need to follow the steps if you've already set up the Marklogic container in the public ui?

Floppy:
I think it's maybe best to point people to the central repo, https://github.com/nationalarchives/ds-caselaw-public-access-service, and put general dev setup instructions there?

Floppy:
Merging this as-is, we can always tweak more later.

PR #23, opened 2022-03-23
View previous versions of a Judgment

lozette:
Show a list of previous versions of a Judgment on the `edit` page.
Any version can be clicked through to, and viewed as normal.

This relies on using v2.1.0 of the API client!

### DIFF
--- a/judgments/views.py
+++ b/judgments/views.py
@@ -15,22 +15,24 @@ from django.utils.translation import gettext
 from lxml import etree
 from requests_toolbelt.multipart import decoder
 
-from judgments.models import Judgment, SearchResult, SearchResults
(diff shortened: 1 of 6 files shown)


judgments/views.py
```diff
@@ -15,22 +15,26 @@
 from lxml import etree
 from requests_toolbelt.multipart import decoder
 
-from judgments.models import Judgment, SearchResult, SearchResults
+from judgments.models import SearchResult, SearchResults
+
+VERSION_REGEX = r"([\d])-([\d]+)"
```
LewisDaleUK, line 20: It doesn't look like this is used anywhere - I think it can be used on line 205 as part of the regex, or just removed completely

Floppy (approved):

PR #32, opened 2022-03-30
Sorted the listed versions by newest first

LornaHarwood:


judgments/views.py
```diff
@@ -197,11 +197,14 @@ def perform_advanced_search(
...
     versions = [
         {
             "uri": part.text.rstrip(".xml"),
-            "version": re.search(r"([\d])-([\d]+).xml", part.text).group(1),
+            "version": int(re.search(r"([\d]+)-([\d]+).xml", part.text).group(1)),
         }
         for part in decoded_versions.parts
     ]
-    return versions
+    sorted_versions = sorted(versions, key=lambda d: d["version"])
```
LewisDaleUK, line 208: if we change the key to:

```python
lambda d: -d["version"]
```

Then we can remove the call to `reverse` in the following line

LewisDaleUK (approved):

PR #81, opened 2022-04-14
add UI for metadata editing

Floppy:


Floppy:
Fantastic, thanks for fixing my half-written code

PR #98, opened 2022-04-21
Explicitly append new elements to the XML tree when updating

lozette:
This PR relies on https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/35
being merged and a new release of the API client

If a metadata element does not exist in the XML document (as opposed to existing
but being empty) we were previously returning `None` from the API client and
the document was therefore uneditable. Now, we will return an empty element if
the element doesn't exist in the document. The element (new or existing) can
then be added to the element tree and re-saved back into the document.

### DIFF
--- a/judgments/views.py
+++ b/judgments/views.py
@@ -21,6 +21,8 @@ from requests_toolbelt.multipart import decoder
 from judgments.models import SearchResult, SearchResults
 
 env = environ.Env()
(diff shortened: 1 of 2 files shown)


Floppy (approved):
Appending an element that's already there has no effect, as shown in https://stackoverflow.com/questions/37572695/python-etree-insert-append-and-subelement; so this is safe as is, and looks good to merge. Worst case, data just doesn't get saved.

PR #101, opened 2022-04-22
Use specific AWS credentials for cloudfront invalidation

Floppy:
I've created a special IAM user that can do it in both environments, and set the appropriate environment variables.


Floppy:
Merging this as I want to check it before the end of the day.

Floppy:
Merging this as I want to check it before the end of the day.

PR #200, opened 2022-06-06
Remove screenreader-only styles to reveal h2 headings in the judgment

lozette:
Copied from e52373ace751969bc36d7e1bdbe29bf8ff7c2b01 in ds-caselaw-public-ui
as this commit is required in this application as well.

https://trello.com/c/tptkA4BN/46-missing-headings-in-html


matt-blair:
Looks good to me. Cross referenced with the same changed on the public UI SASS file for judgment text here: https://github.com/nationalarchives/ds-caselaw-public-ui/commit/e52373ace751969bc36d7e1bdbe29bf8ff7c2b01

PR #233, opened 2022-06-22
Allow editors to edit Neutral Citation of `failure` judgments

lozette:
## Changes in this PR:

Previously, a "red failure" judgment was uneditable in the editor UI. These judgments were ingested but the XML/HTML available was the failure log, not the judgment XML (if it existed). 

After changes to the ingester, we are now attempting to ingest any judgment XML we find, and a judgment will only be "uneditable" if it is a total failure (i.e. it has no judgment XML, just an error log)

An editor can then attempt to edit the judgment, and add a missing neutral citation number. When they do so, **assuming the NCN is in the correct format**, the editor UI will try to create a new URI from the NCN, and move the judgment to that URI.

If this is successful, the editor UI will redirect to the new URI for the judgment and delete the "failure" URI.
If it is not successful, the judgment will remain at the "failure" URI. The editor can then attempt to add a new NCN again, or ask for the judgment to be edited the the clark & reingested. 

https://trello.com/c/8n1Jks1r

### After

Screenshot 2022-06-22 at 11 28 34


judgments/views.py
```diff
@@ -151,9 +172,18 @@ def detail(request):
         context["pdf_url"] = generate_pdf_url(judgment_uri)
 
         if version_uri:
-            context["version"] = re.search(r"([\d])-([\d]+)", version_uri).group(1)
-    except MarklogicResourceNotFoundError:
-        raise Http404(f"Judgment was not found at uri {judgment_uri}")
+            try:
+                version = re.search(r"([\d])-([\d]+)", version_uri).group(1)
```
dragon-dxw, line 176: I think this regex is unnecessarily complicated: we only need the one matching group `(`..`)` for the thing we actually want to match and we don't need `[`..`]` around the `\d` because it's the only character-type we want to look for. (We could say `[\dabcdef]+` for a hexadecimal number).

I suspect we want to match just the first bit of the regex, so this should be `(\d)-\d+`, "the single number before the hyphen followed by at least one digit", or possibly `(\d+)-\d+` if there could be a version number higher than 9.

judgments/views.py
```diff
@@ -151,9 +172,18 @@ def detail(request):
...
+            except AttributeError:
+                try:
+                    version = re.search(
+                        r"([\d]+)-(TDR)-([\d]+)-(.+)", version_uri
```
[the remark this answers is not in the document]
dragon-dxw: Also these are complicated bits of logic that might make sense to pull out into their own function -- isn't this your new extract_version function, in fact?

judgments/views.py
```diff
@@ -289,20 +319,30 @@ def perform_advanced_search(
...
-        for part in decoded_versions.parts
+        for part in decoded_versions
     ]
     sorted_versions = sorted(versions, key=lambda d: -d["version"])
     return sorted_versions
 
 
+def extract_version(version_string: str) -> int:
+    try:
+        return int(re.search(r"([\d]+)-([\d]+).xml", version_string).group(1))
```
dragon-dxw, line 336: See above where I talk about simplifying the regex

judgments/views.py
```diff
@@ -289,20 +319,30 @@ def perform_advanced_search(
...
+    except AttributeError:
+        try:
+            return int(
+                re.search(r"([\d]+)-(TDR)-([\d]+)-(.+).xml", version_string).group(1)
+            )
+        except AttributeError:
+            return 0
```
dragon-dxw, line 343: Is 0 a sensible return value here? It feels potentially misleading. Would None or raising an exception make more/less sense? (You can say int | None in the type string, I believe)

I think:

```
x = re.search()
if not x:
   x = re.search()
return int(x.group(1))
```
might be a slightly cleaner idiom

dragon-dxw (approved):
Mostly fiddly bits -- yell if any of it doesn't make sense.

judgments/views.py
lozette: I'd prefer to leave it as 0 so that the sorting of the versions doesn't throw an exception. In theory we should never get 0, as the versions should match the regex - they don't if they are something like `failures/kittens`. In that case I don't really care if they're all version 0 - something has gone wrong LONG before we got to listing the versions!

dragon-dxw: :+1:

PR #244, opened 2022-06-29
CSS for judgments should be shared between `editor-ui` and `public-ui`

lozette:
## Changes in this PR:

Because `ds-caselaw-public-ui` and `ds-caselaw-editor-ui` are very similar in look and feel, it makes sense that thy should share most of their CSS. 

In the past, when changes were made to public-ui, they would have to be copied across to editor-ui (and sometimes weren't). 

We should treat public-ui as the "base" style and editor-ui as a superset of it, as it contains some functionality that public-ui does not have (specifically, editing judgments).

This PR makes use of Django's staticfiles functionality and git submodules.

1) public-ui is added as a Git submodule. When you check out this project, you will need to run `git submodule init` and `git submodule update` to bring public-ui into this project (see the README). 

2) Django staticfiles copies across all fonts/icons/assets which are shared between both applications

3) Any CSS changes which are for editor-ui *only* should be added to this project. CSS changes which are shared between both projects should be added to `ds-caselaw-public-ui`, merged, and then updated in this project with `git submodule update`

Any duplicated CSS and assets have been deleted in this project. 

https://trello.com/c/OF5DxbU8

### After


gtvj (approved):
Thanks @lozette - this looks good to me (and gave me an opportunity to learn a bit about Git submodules as I've not worked with them before).

PR #248, opened 2022-06-29
Guard against deleted judgments appearing in search results

lozette:
## Changes in this PR:

Something I have noticed on dev/staging but not production - deleted judgments
appearing in the search results. I am not sure if Marklogic caches search
results? 

The effect has been a 500 error on the index page.

In the meantime this change guards against the side effects if this is the case.
If a judgment appears in search results, we try to build a SearchResult
model from it. If we cannot (because the judgment does not exist) a
MarklogicApiError is thrown, which we catch and return None.

If the list of SearchResult objects contains a None, filter it out before
the search results are rendered to the page.

Rollbar: https://rollbar.com/dxw/tna-caselaw-editor-ui/items/50/

### After

### DIFF
--- a/judgments/models.py
+++ b/judgments/models.py
@@ -1,7 +1,9 @@
 # from django.db import models
+import logging
 from datetime import datetime
 from os.path import dirname, join
 
+import caselawclient.Client
 from caselawclient.Client import api_client
 from djxml import xmlmodels
 from lxml import etree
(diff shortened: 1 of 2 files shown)


judgments/models.py
```diff
@@ -81,18 +82,21 @@ def create_from_node(node):
...
+                uri=uri,
+                neutral_citation=judgment.neutral_citation,
+                name=judgment.metadata_name,
+                matches=matches.transform_to_html(),
+                court=judgment.court,
+                date=judgment.date,
+                meta=SearchResultMeta.create_from_uri(uri),
+            )
+        except caselawclient.Client.MarklogicAPIError:
+            return None
```
dragon-dxw, line 99: Could we make a warning that'd hit rollbar?

Not sure if `from warnings import warn` then `warn(e)` is better/worse than targetting Rollbar directly with     `rollbar.report_message(e, 'warning')` or something.

dragon-dxw (approved):

judgments/models.py
lozette: I've added `logging.warning` as we use that in public-ui already. I'm not sure of the difference between that and `warnings`!

dragon-dxw: No, that's better. (Warnings I think just dumps it to STDERR, and is basic Python functionality.) ✅

PR #255, opened 2022-07-04
Copy assets when a judgment is moved in Marklogic to a new URI

lozette:
## Changes in this PR:

When a judgment is moved from a "failure" URI (e.g. `failures/TDR-2022-DNWR`) to
an NCN-derived URI (e.g. `ukpc/2023/120`) we want to move the assets (docx,
images, pdf etc) to an S3 bucket location that matches the URI.

This enables the assets to be downloadable from the editor UI, as the view code
looks for the assets on S3 in a bucket location that matches the URI. E.g. the assets 
for `ukpc/2023/120` will be in S3 at `{bucket}/ukpc/2023/120`

https://trello.com/c/cy2ZElXN

### After


judgments/utils.py
```diff
@@ -78,3 +78,14 @@ def set_metadata(old_uri, new_uri):
     ]:
         if value is not None:
             api_client.set_property(new_uri, key, value)
+
+
+def build_new_key(old_key, new_uri):
+    old_filename = old_key.rsplit("/", 1)[-1]
+
+    if "docx" in old_filename or "pdf" in old_filename:
```
dragon-dxw, line 112: I'd prefer `old_filename.endswith(".docx")` (and "`.pdf`") -- just in case!

judgments/utils.py
```diff
@@ -78,3 +78,14 @@ def set_metadata(old_uri, new_uri):
...
+        new_filename = new_uri.replace("/", "_")
+        return f"{new_uri}/{new_filename}.{old_filename.rsplit('.', 1)[-1]}"
```
dragon-dxw, line 114: This feels a little overcompliated: why not `old_filename.split('.')[-1]`? (I suppose yours is a little more algorithmically efficient...)

judgments/utils.py
```diff
@@ -79,6 +79,10 @@ def set_metadata(old_uri, new_uri):
         if value is not None:
             api_client.set_property(new_uri, key, value)
 
+    published = api_client.get_published(old_uri)
+    api_client.set_boolean_property(new_uri, "published", bool(published))
```
dragon-dxw, line 93: > Technically this should always be "false" but copy it over anyway, just to be 100% sure.

This feels like something an inline comment would help readability for.

judgments/utils.py
```diff
@@ -83,6 +93,18 @@ def set_metadata(old_uri, new_uri):
...
+    client = create_s3_client()
+    bucket = env("PRIVATE_ASSET_BUCKET")
+    response = client.list_objects(Bucket=bucket, Prefix=old_uri)
+
+    for result in response.get("Contents", []):
+        old_key = str(result["Key"])
+        new_key = build_new_key(old_key, new_uri.lstrip("/"))
+        if new_key is not None:
+            source = {"Bucket": bucket, "Key": old_key}
+            client.copy(source, bucket, new_key)
```
dragon-dxw, line 106: I feel like I'd maybe like to see some wide-spectrum error handling here, if only to make sure that the erroring filenames are properly output before potentially re-raising the error.

Not essential though.

dragon-dxw (approved):

judgments/utils.py
lozette: I have no idea, it's probably what came up when I googled!

judgments/utils.py
lozette: Yeah it does seem like we do a lot of copying files around (in the ingester as well as editor) without handling any potential errors.

PR #263, opened 2022-07-06
(Fix) When moving & retreiving items in S3, strip the leading slash from keys

lozette:
## Changes in this PR:

While debugging the functionality to move assets around in S3 when a judgment is
moved, we found that the S3 moving & retrieval functions are very sensitive
to a leading slash on the uri. For example looking for a prefix of `/xxx/2022`
in S3 returns nothing but `xxx/2022` returns items.

Additionally, the docx/pdf URL generation will look for a filename with a leading
underscore if a leading slash is on the URI.

To mitigate this, strip the URI of a leading slash whenever we use it to
find items in S3.

https://trello.com/c/cy2ZElXN

### After

### DIFF
--- a/judgments/utils.py
+++ b/judgments/utils.py
@@ -101,10 +101,14 @@ def set_metadata(old_uri, new_uri):
 def copy_assets(old_uri, new_uri):
     client = create_s3_client()
     bucket = env("PRIVATE_ASSET_BUCKET")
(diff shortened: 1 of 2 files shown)


dragon-dxw (approved):
There's a part of me that's concerned that we might have code that passes those URIs around with and without slashes and we should probably document that one or the other is the correct canonical format to pass around...

PR #355, opened 2022-09-02
CSS for judgments should be shared between editor and public UIs

lozette:
## Changes in this PR:

The Judgment display CSS `_judgment_text.scss` should be the same between both
the editor and public UI appls. Ensuring edits to the public UI were being
replicated to the editor UI was tricky as it replied on the developers
remembering to make changes in both places.

Instead, we should share the judgment CSS between both apps. Public UI is the
"source of truth".

In the Dockerfile, use the Github API to get the latest tag for the public UI.
Then use that tag to build a URL and retrieve the judgment CSS using curl.
Finally, copy the judgment CSS to the correct place in the editor UI.

`Dockerfile` is the Dockerfile used in production environments.
`compose/local/django/Dockerfile` is the Dockerfile used be Fab for local
development.

https://trello.com/c/OF5DxbU8/603-css-for-judgments-should-be-shared-between-editor-and-public-uis

### After


rjw1 (approved):
Looks fine. ship it. At worst staging wont deploy and i can look at it then

PR #369, opened 2022-09-12
Add CSS sharing to `fab run`

lozette:
## Changes in this PR:

In https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/355 we added shared CSS from ds-caselaw-public-ui to this repository via the Dockerfile.

However, some of the developers found that they were unable to see the shared CSS when running the application with `fab run`. `fab build` and `fab start` build and run the Docker containers using the Dockerfile and retrieve the latest version of the CSS every time they are run. But `fab run` is more "lightweight" and does not retrieve the latest CSS every time.

Amend `fab run` to explicitly get the CSS from public-ui, so that developers using this command to run the application always get the most recent version of the shared CSS.

### After

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -75,6 +75,15 @@ def run(c):
     django_exec("python manage.py migrate")
     django_exec("rm -rf /app/staticfiles")
     django_exec("python manage.py collectstatic")
+    # Get the public-ui version of _judgment_text.scss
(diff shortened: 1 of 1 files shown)


timcowlishaw:
Aha ace, thank you Laura! That seems like a sensible solution!

Floppy (approved):

PR #370, opened 2022-09-12
Chore/setup tweaks from tim

timcowlishaw:
## Changes in this PR:
Fixes a few stumbling blocks when starting up the project for the first time:

- Ensures the correct version of `sass` is run in the `npm` scripts
- Ensures the `_judgements_text.scss` SASS is not committed to this repo (the canonical version is included from [ds-caselaw-public-ui](https://github.com/nationalarchives/ds-caselaw-public-ui)
- Adds a note that [pre-commit](https://pre-commit.com/) is needed to run the pre-push hook.


lozette (approved):
Approved but needs linting!

PR #387, opened 2022-09-21
Fix/visited links

Terry-Price:
## Changes in this PR:
Fixed the issue that users could not easily identify what links they have already visited. So I have created a visited colour and reordered the pseudo links CSS so they are in the correct order.

### Before
Screenshot 2022-09-21 at 10 41 29

### After
Screenshot 2022-09-21 at 10 41 58


Terry-Price:
Hi Laura,
Thanks so much for adding me to the EUI permissions.
I can now make changes which is great.
I hope I have done everything okay, shout if not.

lozette (approved):

PR #405, opened 2022-09-29
Use elements from the search results XML to build SearchResult models

lozette:
## Changes in this PR:

Previously, we were making a call to the Marklogic API for every search result, using the retrieved XML to find the name, date, citation and court of each result. This meant we had at least 10 calls to the API on each page (as each page contains at least 10 search results).

Instead, we have now added this data to the search results XML directly from Marklogic. We can use Xpath to pick the required data out from the search results themselves, and lose the extra API calls.

In addition, we can also remove the Judgment model. Its only purpose was to hold the data required for the SearchResult model, and it is no longer needed. 

https://trello.com/c/wBZUsMuz

### After


judgments/models.py
```diff
@@ -79,21 +60,44 @@ def __init__(
...
     @staticmethod
     def create_from_node(node):
+        namespaces = {
+            "search": "http://marklogic.com/appservices/search",
+            "uk": "https://caselaw.nationalarchives.gov.uk/akn",
+            "akn": "http://docs.oasis-open.org/legaldocml/ns/akn/3.0",
+        }
         uri = node.xpath("@uri")[0].lstrip("/").split(".xml")[0]
+        neutral_citation = (
+            node.xpath("search:extracted/uk:cite", namespaces=namespaces)[0].text
```
dragon-dxw, line 70: Did you decide against `neutral_citation = cite[0].text if cite := node.xpath('...') else ""` ?

dragon-dxw (approved):

judgments/models.py
lozette: I couldn't get it to work, not sure if I was doing it wrong:
```
neutral_citation = cite[0].text if cite := node.xpath("search:extracted/uk:cite", namespaces=namespaces) else ""
```
```
File "/app/judgments/models.py", line 74
    neutral_citation = cite[0].text if cite := node.xpath("search:extracted/uk:cite", namespaces=namespaces) else ""
                                            ^
SyntaxError: invalid syntax
```

dragon-dxw: Might need brackets around it: `... if (cite := ...) else ...`

PR #424, opened 2022-10-07
Fix: Strip whitespace from around NCN

lozette:
Trello: https://trello.com/c/cXbWogEp/1053-issue-editing-a-judgment

An editor was having trouble editing the NCN for a failed judgment. It appears the NCN they were pasting into the form had a trailing space, which was not being stripped, and was causing the call to ds_caselaw_utils.neutral_uri() to fail.

Trim any leading or trailing whitespace from the NCN before sending it to ds_caselaw_utils.neutral_uri()

### After

### DIFF
--- a/judgments/tests.py
+++ b/judgments/tests.py
@@ -260,6 +260,25 @@ class TestUtils(TestCase):
         assert result == "new/uri"
 
     @patch("judgments.utils.api_client")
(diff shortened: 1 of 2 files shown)


dragon-dxw (approved):
:+1:

PR #431, opened 2022-10-12
Change Subject of SNS notification message

lozette:
The SNS Subject must begin with an ASCII character and be under a certain char limit; long judgment names make the subject too long. Replace the judgment name with the URI in the SNS message subject.

### After


dragon-dxw (approved):
Thank you!

PR #434, opened 2022-10-17
Document how to delete judgments

lozette:
Until there's a delete button in the editor UI for judgmetns with an NCN, we need to manually delete them. This documents the process.

We will have a delete button in the Editor v2.0.


dragon-dxw (approved):
Thanks!

PR #448, opened 2022-10-24
Allow XML download of judgment

lozette:
## Changes in this PR:

A quick button/hackable URL to enable download of the XML for a judgment from the editor.

This can be made nicer in Editor2.0!!

https://trello.com/c/5YX4JwZ9/1106-add-ability-to-see-xml-in-editor-site-editorfindcaselaw

Screenshot 2022-10-24 at 16 03 54

### DIFF
--- a/judgments/views.py
+++ b/judgments/views.py
@@ -188,6 +188,7 @@ def detail(request):
     judgment_uri = params.get("judgment_uri", None)
     version_uri = params.get("version_uri", None)
     context = {"judgment_uri": judgment_uri, "is_failure": False}
+
     try:
         judgment_xml = api_client.get_judgment_xml(judgment_uri, show_unpublished=True)
         judgment_root = get_judgment_root(judgment_xml)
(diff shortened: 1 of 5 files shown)


judgments/views.py
```diff
@@ -188,6 +188,15 @@ def detail(request):
     judgment_uri = params.get("judgment_uri", None)
     version_uri = params.get("version_uri", None)
     context = {"judgment_uri": judgment_uri, "is_failure": False}
+
+    if judgment_uri.endswith(".xml"):
```
dragon-dxw, line 192: I feel like this should be a different route entirely? `/xmlview?judgment_uri=ewhc/fam/2022/1` perhaps? That there's only three lines of shared context and a lot more unshared feels like this should be a different function.

judgments/views.py
```diff
@@ -224,6 +225,18 @@ def detail(request):
     return HttpResponse(template.render({"context": context}, request))
 
 
+def detail_xml(request):
+    params = request.GET
+    judgment_uri = params.get("judgment_uri", None)
+    try:
+        judgment_xml = api_client.get_judgment_xml(judgment_uri, show_unpublished=True)
```
dragon-dxw, line 232: Given this line is the only one that can raise the MarkLogicResourceNotFound error, I'd typically put the exception immediately after it. Doesn't have to change.

dragon-dxw (approved):

PR #454, opened 2022-10-26
Assign judgment to self from index page

dragon-dxw:
TODO: say who judgment is assigned to on search page.

### After


dragon-dxw:
[👩‍💻EUI - Assign Judgments HTTP endpoint](https://trello.com/c/m8KuGSEk/33-%F0%9F%91%A9%F0%9F%92%BBeui-assign-judgments-http-endpoint)

PR #455, opened 2022-10-27
Editor priorities selection

dragon-dxw:
Doesn't feature actually showing the priority. Sorting is next.

### After


dragon-dxw:
[👩‍💻EUI - Judgment Priorities](https://trello.com/c/BKcvFpu3/34-%F0%9F%91%A9%F0%9F%92%BBeui-judgment-priorities)

lozette (approved):

judgments/views.py
```diff
@@ -279,7 +281,10 @@ def parse_priority(priority: Union[str, int]) -> Optional[str]:
     if priority:
         api_client.set_property(judgment_uri, "editor-priority", priority)
         target_uri = request.META.get("HTTP_REFERER") or "/"
+
+        messages.success(request, "Judgment priority set.")
```
lozette, line 285: I don't know if it's worth adding the judgment URI/name and priority to the message? Although the editor will be able to see it on the page.

PR #456, opened 2022-10-28
Bump version of caselaw-utils to include new tribunals

timcowlishaw:
## Changes in this PR:

Fairly self explanatory - this fixes the filters for upper tier tribunals and brings in the new lower tier tribunals to the list and filters


Terry-Price (approved):
all good

PR #500, opened 2022-11-22
Add a check for secrets

jacksonj04:
## Changes in this PR:

This action will check the repository for committed secrets, and fail (allowing the user to take action) if they're present.

https://trello.com/c/VIcWTQjM/139-add-check-for-secrets-to-ds-caselaw-editor-ui


.gitallowed
```diff
@@ -0,0 +1 @@
+000000000000
```
dragon-dxw, line 1: What is this tripping up on?

jacksonj04: It's a dummy queue ID as used by `localstack.sh`.

PR #504, opened 2022-11-23
FCL-15 Ensure there is always a string for the judgment

dragon-dxw:
Whilst this doesn't fix the underlying issue, and is technically wrong (there is a title for the judgment, it's just not coming through to the front end in this context for some reason, probably related to how the Custom API's request to MarkLogic), having a placeholder here means the editor has a chance to be able to click a link to visit the judgment and make changes.

There are legitimately unpublished judgments which may have "" as a title, so this is a useful change.

image

[note also that the search box is blank and it shouldn't be... possibly related...]


jacksonj04 (approved):
This looks like a solid fallback to me, :shipit:

jacksonj04 (approved):

PR #521, opened 2022-12-02
Add a standard header to the README

jacksonj04:
This will help orient people viewing the code within the larger project.


dragon-dxw (approved):
Not 100% sure about "*archived* case law" -- it somewhat implies that there's a distinction in what the editors can handle.

PR #562, opened 2022-12-19
Fix bug where assignment field empty after editing judgment

timcowlishaw:
## Changes in this PR:

This was caused by the fact that the dictionary of user names and ids was not populated in the handler for the POST request, so the dropdown was empty on rendering the template. Fixed by factoring the code to generate this out into a helper method, then calling it from both the POST and GET handler for judgment edits.

https://trello.com/c/fIcztXy5/93-%F0%9F%91%A9%F0%9F%92%BB%F0%9F%90%9E-after-submitting-change-to-judgment-assigned-to-is-visually-empty


jacksonj04 (approved):
🚢

PR #585, opened 2023-01-04
Factor out referrer parsing and add check for redirect spoofing

timcowlishaw:
## Changes in this PR:

adds a helper method to check that the referrer url is local and conditionally return for redirects. This stops people spoofing the referrer, and (hopefully) silences the security warnings we're getting from CodeQL.

### DIFF
--- a/judgments/utils/__init__.py
+++ b/judgments/utils/__init__.py
@@ -1,6 +1,7 @@
 import re
 import xml.etree.ElementTree as ET
 from datetime import datetime
+from urllib.parse import urlparse
 
 import ds_caselaw_utils as caselawutils
 from caselawclient.Client import (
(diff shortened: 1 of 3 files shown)


timcowlishaw:
D'you know what - now this is extracted out, I should really add some unit tests. I'll do that now!

judgments/utils/__init__.py
```diff
@@ -25,6 +26,20 @@ class NeutralCitationToUriError(Exception):
     pass
 
 
+def is_url_relative(url):
+    return not bool(urlparse(url).netloc)
+
+
+def referrer_url(request, default="/"):
```
dragon-dxw, line 33: Suggestion: rename to `ensure_local_referrer_url` and add docstring, something like:
```python
"""
Make sure that we do not redirect the user to a website we do not control.
In future we should explicitly specify a return URL in the POST data when clicking a button,
rather than parsing the HTTP_REFERER header.
"""
```

dragon-dxw (approved):

PR #594, opened 2023-01-05
Force locale to be en-gb

dragon-dxw:
If the browser locale is set to a locale other than `en-gb`, untranslated strings appear (e.g. `judgments.details` instead of `Judgment details:`.

We state explicitly that `en-gb` is the only locale available so everything will fall back to it.

image
image


jacksonj04 (approved):
🇬🇧

PR #599, opened 2023-01-12
Don't create blank uk:court tag if no court provided

dragon-dxw:
Having completed this, I've just realised I'm fixing it in the wrong spot -- should probably fix it in set_judgment_court in the api client...

This might be worth having anyway.

### DIFF
--- a/judgments/tests.py
+++ b/judgments/tests.py
@@ -411,6 +411,42 @@ class TestJudgmentEditor(TestCase):
         mock_api.get_property.assert_called_with("ewhc/ch/1999/1", "assigned-to")
         assert b"selected>otheruser" in response.content
 
+    @patch("judgments.views.edit_judgment.api_client")
+    @patch("judgments.views.edit_judgment.unpublish_documents")
+    @patch("judgments.views.edit_judgment.notify_changed")
+    @patch("judgments.views.edit_judgment.EditJudgmentView.get_metadata")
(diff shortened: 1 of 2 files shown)


judgments/tests.py
```diff
@@ -411,6 +411,42 @@ def test_assigned(self, mock_api):
         mock_api.get_property.assert_called_with("ewhc/ch/1999/1", "assigned-to")
         assert b"selected>otheruser" in response.content
 
+    @patch("judgments.views.edit_judgment.api_client")
+    @patch("judgments.views.edit_judgment.unpublish_documents")
+    @patch("judgments.views.edit_judgment.notify_changed")
+    @patch("judgments.views.edit_judgment.EditJudgmentView.get_metadata")
+    def test_no_court_no_problem(self, mock_metadata, mock_sns, mock_s3, mock_api):
```
jacksonj04, line 418: > That was _T. N. A._ with their latest hit _No Court, No Problem_...

Could this be something like `test_court_not_set_if_not_specified, just for extra clarity?

jacksonj04 (changes requested):
One smol naming tweak, but happy with this. Agree there should be a matching bit on the API going "you can't set a null court".

dragon-dxw:
Me and Anthony have agreed this is wrong and only the Marklogic API Client should handle this -- a request to set an empty court should delete the court tag and that's not the editor-ui's responsibility.

PR #601, opened 2023-01-17
Clarify machine name issues with VPN.

dragon-dxw:
Clarify names of machines vary for TNA/dxw.

### After


jacksonj04 (approved):

anthonyhashemi:
I have created a PR which attempts to simplify the staging marklogic section and addresses this in it. Worth checking that first: https://github.com/nationalarchives/ds-caselaw-public-ui/pull/525

PR #608, opened 2023-01-23
Judgment toolbar refinements

jacksonj04:
A series of adjustments (mostly visual) to the judgment toolbar:

- Improvements to conditional wrapping to eliminate an empty (layout changing) element, fixing unbalanced whitespace and further reducing header height
- Migrate the "download XML" link to the sidebar of the judgment edit page, since it's not part of the key workflow
- Fix adding context to judgment content view, so the 'view on public site' button correctly displays
- Make all toolbar buttons display in all contexts, even if disabled, so that toolbar buttons in the same place always do the same thing
- Subtle hover effect on buttons helps highlight current interaction state
- The button for the current view is highlighted differently
- Changes to language in toolbar buttons makes intent clearer, improves consistency, reduces overall length and removes unnecessary repetition of context

### Before

localhost_3000_edit_judgment_uri=ewhc_admin_2023_81 (1)

### After

localhost_3000_edit_judgment_uri=ewhc_admin_2023_81

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_judgment_text_toolbar.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_judgment_text_toolbar.scss
@@ -5,15 +5,33 @@
   padding: $spacer__unit;
   border-bottom: 0.125rem solid $color__grey;
 
-  a {
+  &__button {
     background-color: $color__aqua-blue;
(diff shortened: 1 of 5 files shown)


timcowlishaw (approved):
👏 👏 👏

PR #623, opened 2023-01-30
Add submitter details to Jira magic link

jacksonj04:
Editors usually copy a submitter's details into a Jira issue. Save them from doing this by automatically adding it to the issue body.

### DIFF
--- a/judgments/views/edit_judgment.py
+++ b/judgments/views/edit_judgment.py
@@ -11,6 +11,7 @@ from django.http import Http404, HttpResponse
 from django.shortcuts import redirect
 from django.template import loader
 from django.urls import reverse
+from django.utils.translation import gettext
 from django.views.generic import View
 from requests_toolbelt.multipart import decoder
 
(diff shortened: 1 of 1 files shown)


judgments/views/edit_judgment.py
```diff
@@ -92,7 +93,19 @@ def build_jira_create_link(self, request, context):
...
 
         description_string = "{editor_details_url}".format(
-            editor_details_url=editor_details_url
+            editor_details_url="""{details_url}
+
+{source_name_label} {source_name}
+{source_email_label} {source_email}
+{consignment_ref_label} {consignment_ref}""".format(
+                details_url=editor_details_url,
+                source_name_label=gettext("judgments.submitter"),
```
dragon-dxw, line 102: I think it's idiomatic to use `_` as the spelling of `gettext`, but not something that needs changing.

dragon-dxw (approved):

judgments/views/edit_judgment.py
jacksonj04: The rest of the application uses `gettext`, so I'll err with consistency.

PR #664, opened 2023-02-09
Update contents of publication confirmation email

jacksonj04:
A series of tweaks to the confirmation of publication email body which is generated for editors.

### DIFF
--- a/ds_caselaw_editor_ui/templates/emails/confirmation_to_submitter.txt
+++ b/ds_caselaw_editor_ui/templates/emails/confirmation_to_submitter.txt
@@ -9,4 +9,4 @@ We cannot make any amendments to the content or formatting of the judgment. If y
 If you need to contact us regarding your judgment please email judgments@nationalarchives.gov.uk quoting your case details and TDR reference "{{ reference }}".
 
 Kind regards,
-XXXXXX
+{{ user_signature }}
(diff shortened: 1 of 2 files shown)


ds_caselaw_editor_ui/templates/emails/confirmation_to_submitter.txt
```diff
@@ -1,7 +1,12 @@
-We are pleased to inform you that your judgment reference {{ reference }} has been published on Find Case Law, and can be viewed at {{ public_judgment_url }}.
+Dear XXXXXX,
```
dragon-dxw, line 1: I'd have been tempted to omit the Dear line, so the email is complete.
There's a "Contact-Name": "Miles Morales" field in the TDR metadata which agrees with the "Contact-Email" field; might be scope for future improvement.

dragon-dxw (approved):

ds_caselaw_editor_ui/templates/emails/confirmation_to_submitter.txt
jacksonj04: That's what we went with originally, but feedback from editors was they would prefer it with a placeholder ready to fill.

Using the name from TDR would be nice... but turning a full name string into a form of address is sketchy at best.

PR #669, opened 2023-02-10
Refactor Judgment read interactions in detail/edit views to a single model

jacksonj04:
There is a lot of code in the judgment views which handles getting information about a judgment. This is ripe territory for duplication of concerns, as well as being awkward to test and isolate the behaviour of.

This refactoring PR moves those 'get information' functions into a new Judgment model class, simplifying interactions with judgments.

CodeClimate issues are all actually pre-existing issues, but because I've touched bits of the file they're flagging as new ones.

## Future

- We should move write operations to this class as well.
- This class should give us a nice wedge into improving how we provide mocks for testing purposes.
- When this model can handle all interactions we should move it to being a function of the api-client library, and begin migrating other actors to use it instead of talking to the API directly.


dragon-dxw:
There's an unverified commit you'll probably want to sign  (rebase interactively, set to `edit`, `git commit -S '-s'` seems to work for me)

jacksonj04:
@dragon-dxw how odd! Rebased and fixed all the signatures.

judgments/views/edit_judgment.py
```diff
@@ -166,8 +119,12 @@ def render(self, request, context):
     def get(self, request, *args, **kwargs):
         params = request.GET
         judgment_uri = params.get("judgment_uri")
+        judgment = Judgment.objects.get_by_uri(judgment_uri)
+
         context = {"judgment_uri": judgment_uri}
-        context.update(self.get_metadata(judgment_uri))
```
anthonyhashemi, line 170: so previously we were getting all metadata to update it here but actually we only needed judgement and page title as the others werent changing anyway I'm guessing

judgments/models.py
```diff
@@ -154,3 +160,115 @@ class Meta:
         namespaces = {"search": "http://marklogic.com/appservices/search"}
 
     transform_to_html = xmlmodels.XsltField(join(dirname(__file__), "search_match.xsl"))
+
+
+class JudgmentManager:
+    @classmethod
+    def get_by_uri(cls, uri: str) -> "Judgment":
```
anthonyhashemi, line 167: Just so I understand, we have the Judgement class which is not actually a django model right? It looks like a pseudo model which has an objects property which gives access to the judgementmanager.

And then within the properties of the class, we use the api_client to access stuff from marklogic.

Is this a standard pattern that we use when trying to abstract away logic surrounding a specific entity coming from marklogic?

judgments/views/edit_judgment.py
jacksonj04: That's correct - everything which `get_metadata()` made available via a dict (which was then merged into `context`) is now available via `context.judgment` via the instance of the `Judgment` model class.

judgments/models.py
jacksonj04: That's correct - it's not a real Django model, but I've tried to make it behave enough like one (eg by using the `objects` property to refer to the manager class) that it should feel kinda right.

It's not a standard pattern we use, but it's a pattern we _should_ use so that controllers aren't having to interact directly with the API, and instead all interaction happens through the `Judgment` model. Eventually I'd like the `api_client` to spit back this same model and everything uses it to interact, so there's no direct coupling at all to an understanding of what goes on at the API or MarkLogic, but that's down the road.

judgments/views/edit_judgment.py
anthonyhashemi: ah of course

judgments/models.py
anthonyhashemi: Yeah abstracting it away from the controller is definitely the way forward. 

I do have a slight concern that intermingling the idea of django models and these types of pseudo models _could_ lead to confusion.

If agreed that it is clearer to seprate those ideas, then we could just leave this as a class but instead of the JudgementManager just have the classmethod on the Judgement class like a standard factory method on a class.


Another thing that might be useful, to separate concerns of editor-ui and marklogic client, would be to pass in the api_client as an object to the class.

Putting that together, what do you think of:

```
class Judgment:
    def __init__(self, uri: str, api_client):
        self.uri = uri
        self.api_client = api_client

 @classmethod
    def get_by_uri(cls, uri: str) -> "Judgment":
         return Judgment(uri)

    @cached_property
    def neutral_citation(self) -> str:
        return self.api_client.get_judgment_citation(self.uri)
....
```

judgments/models.py
```diff
@@ -154,3 +160,115 @@ class Meta:
...
+        return Judgment(uri)
+
+
+class Judgment:
```
anthonyhashemi, line 171: `Judgement`

judgments/models.py
```diff
@@ -154,3 +162,118 @@ class Meta:
         namespaces = {"search": "http://marklogic.com/appservices/search"}
 
     transform_to_html = xmlmodels.XsltField(join(dirname(__file__), "search_match.xsl"))
+
+
+class Judgment:
```
anthonyhashemi, line 167: need to replace `judgment` with `judgement` throughout the PR

jacksonj04: @anthonyhashemi it turns out that whilst "judgement" is the ostensibly correct version in UK English, the legal profession still prefers "judgment".

Caught me out for the first few weeks.

judgments/tests/test_models_judgment.py
```diff
@@ -0,0 +1,157 @@
...
+
+from judgments.models import Judgment
+
+
+@pytest.fixture
+def mock_api_client():
+    return Mock(spec=MarklogicApiClient)
+
+
+class TestJudgment:
```
anthonyhashemi, line 15: since we have set the default value for api_client, can we add a unit test for when we instantiate with only uri?

```
def test_judgement_with_default_api_client():
    judgement = Judgement(uri)
    assert isinstance(judgement.api_client, MarklogicApiClient)
    assert judgement.api_client.host == settings.MARKLOGIC_HOST
    assert judgement.api_client.username == settings.MARKLOGIC_USER
    assert judgement.api_client.password == settings.MARKLOGIC_PASSWORD
    assert judgement.api_client.use_https == settings.MARKLOGIC_USE_HTTPS
```

alternatively could remove the default - whichever you prefer

judgments/models.py
anthonyhashemi: seriously! just looks so odd to me - my bad 😄

anthonyhashemi:
Looks good!

anthonyhashemi (approved):

PR #678, opened 2023-02-14
Allow editors to break checked out (locked) documents.

dragon-dxw:
When the enrichment pipeline was broken, it was locking a lot of documents. Editors need to be able to unlock these documents in case they need to be able to unpublish etc. in an emergency.

Could do with a front end pass.

Has a confirmation screen but does not address:

> how is this audited, should someone receive a message to say something has been deleted?

(I think that this is a relatively benign operation -- it could be used to block enrichment, but that's not a major concern, unlike deletion)


dragon-dxw:
[EUI: Document enrichment lock (override button?)](https://trello.com/c/dTLlGGsY/419-eui-document-enrichment-lock-override-button)

dragon-dxw:
Needs link from edit page.

dragon-dxw:
image

ds_caselaw_editor_ui/templates/judgment/confirm-unlock.html
```diff
@@ -0,0 +1,21 @@
...
+  </header>
+{% endblock header %}
+{% block content %}
+  <div class="unlock-judgment">
+    <p>{{context.judgment_uri}}</p>
+    <p class="unlock-judgment__confirmation">{% translate "judgment.confirm_unlock" %}</p>
+    <form method="POST">
+      {% csrf_token %}
+      <input type="hidden" name="judgment_uri" value="{{context.judgment_uri}}">
+      <button type="submit">unlock</button>
```
jacksonj04, line 17: Should we translate this string?

jacksonj04 (commented):
One little "should we translate" query, other than that this looks good to me.

ds_caselaw_editor_ui/templates/judgment/confirm-unlock.html
dragon-dxw: Done: now reads Unlock judgment

Terry-Price (approved):
looks good to me Dragon. We'll put this in backlog and relook soon.
Happy to merge as it stands though.

PR #690, opened 2023-02-20
Do not run CodeQL in merge queue commits

jacksonj04:
CodeQL does not reliably run in merge queue commits since it's not a required check (so commits may be automatically removed after merge but before CodeQL has reported).


dragon-dxw:
I can't see any non-linty changes -- is that correct?

dragon-dxw (approved):

PR #700, opened 2023-02-21
Improve coverage of mypy checks, and add mypy to CI

jacksonj04:
At the moment mypy under `fab test` is only checking the `ds_caselaw_editor_ui` folder, but we also implement a lot of stuff in the `judgments` folder and this isn't being analysed.

In addition, mypy isn't being run at all in CI. Add a new task to the workflow to run it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,10 +35,29 @@ jobs:
       - name: Run pre-commit
         uses: pre-commit/action@v3.0.0
 
+  static-analysis:
+    runs-on: ubuntu-latest
+    needs: linter
+    steps:
+      - name: Checkout Code Repository
+        uses: actions/checkout@v3
+
+      - name: Copy .env.example file
+        uses: canastro/copy-file-action@master
+        with:
+          source: ".env.example"
+          target: ".env"
+
+      - name: Build the Stack
+        run: docker-compose build django postgres
+
+      - name: Run mypy
+        run: docker-compose run django mypy ds_caselaw_editor_ui judgments
+
   # With no caching at all the entire ci process takes 4m 30s to complete!
   pytest:
     runs-on: ubuntu-latest
-
+    needs: linter
     steps:
       - name: Checkout Code Repository
         uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -35,6 +35,24 @@ jobs:
...
+      - name: Copy .env.example file
+        uses: canastro/copy-file-action@master
+        with:
+          source: ".env.example"
+          target: ".env"
+
+      - name: Build the Stack
+        run: docker-compose build postgres django
+
+      - name: Run mypy
```
anthonyhashemi, line 53: do we need to build the stack to run mypy?

jacksonj04: Annoyingly, yes. I'm debating moving it to just a step in the regular tests so we're not building it twice.

dragon-dxw:
Was going to suggest adding mypy to pre-commit, but it's a pain. Got as far as

 ```yaml
  - repo: https://github.com/pre-commit/mirrors-mypy
    rev: v1.0.1
    hooks:
      - id: mypy
        additional_dependencies:
          - django-stubs
          - django
          - djangorestframework
          - django-environ
      ```
      
      before giving up because it's running the code and complaining about environment variables.

jacksonj04:
@dragon-dxw that's _exactly_ where I got to before giving up as well!

judgments/utils/__init__.py
```diff
@@ -134,10 +134,8 @@ def render_versions(decoded_versions):
 
 
 def extract_version(version_string: str) -> int:
-    try:
-        return int(re.search(VERSION_REGEX, version_string).group(1))
-    except AttributeError:
-        return 0
+    result = re.search(VERSION_REGEX, version_string)
```
dragon-dxw, line 137: I have no idea what this code was meant to be doing.

dragon-dxw (approved):
I'm wary that if Mypy is only running in CI, it's going to be a pain to fix Mypy problems. Would very much like a pre-commit hook.

I hope it is worth the extra pain :)

judgments/utils/__init__.py
jacksonj04: it looks like it was supposed to handle the error case of "there are no matches" by returning `0`.

PR #706, opened 2023-02-23
Swap editing a judgment's metadata to use POST-Redirect-GET pattern

jacksonj04:
At the moment, whenever changes are made to a judgment the view will perform the changes and then directly return a `HTTP 200` with the edit form. This has led to some problems with missing template context, eg judgment details being unavailable for magic links.

By swapping to a [POST-Redirect-GET](https://en.wikipedia.org/wiki/Post/Redirect/Get) pattern, this view will perform the changes and then redirect the user back to the usual GET form view.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -4,6 +4,7 @@ Base settings to build other settings files upon.
 from pathlib import Path
 
 import environ
+from django.contrib.messages import constants as message_constants
 
 ROOT_DIR = Path(__file__).resolve(strict=True).parent.parent.parent
 # ds_judgments_editor_ui/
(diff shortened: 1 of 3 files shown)


config/settings/base.py
```diff
@@ -291,3 +292,8 @@
 )
 
 JIRA_INSTANCE = env("JIRA_INSTANCE")
+
```
dragon-dxw, line 295: https://docs.djangoproject.com/en/4.1/ref/contrib/messages/#message-tags suggests this is absolutely standard. Today I learned! :)

dragon-dxw (approved):

PR #711, opened 2023-02-23
Improve robustness of generating public URIs

jacksonj04:
It's possible to provide a judgment URI with trailing or leading slashes. This doesn't affect the behaviour of retrieving the record, but _does_ make it look unexpected when navigating around or sharing links.

This PR adds behaviour where the `Judgment` class will strip those slashes on the way in. It also centralises behaviour for rendering the public URI of a judgment into that class, so we have less to maintain.

### DIFF
--- a/judgments/tests/test_models_judgment.py
+++ b/judgments/tests/test_models_judgment.py
@@ -29,6 +29,15 @@ class TestJudgment:
 
         mock_api_client_class.assert_not_called()
 
+    def test_public_uri(self, mock_api_client):
+        mock_api_client.get_judgment_citation.return_value = "2023/test/1234"
+
+        judgment = Judgment("test/1234", mock_api_client)
+
+        assert (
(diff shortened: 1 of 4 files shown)


judgments/tests/test_models_judgment.py
```diff
@@ -29,6 +29,20 @@ def test_judgment_doesnt_use_default_api_client_if_provided(
 
         mock_api_client_class.assert_not_called()
 
+    def test_uri_strips_slashes(self, mock_api_client):
+        judgment = Judgment("////test/1234/////", mock_api_client)
+
+        assert judgment.uri == "test/1234"
+
+    def test_public_uri(self, mock_api_client):
+        mock_api_client.get_judgment_citation.return_value = "2023/test/1234"
```
dragon-dxw, line 38: this test feels very odd --

a citation probably looks like "[2023] TEST 1234", I think?
are we showing that it's a substring of the judgment citation? Or that it's the value taken from the new Judgment call?

dragon-dxw (approved):

PR #713, opened 2023-02-24
Add feature flag to remove the priority in EUI

Donna-H:
## Changes in this PR: 
Add feature flag to remove the priority function
https://trello.com/c/YnFDTZ56/494-3-eui-live-experimental-get-rid-of-priority-and-see-what-happens
### Before
before-priority

### After
after-remove-priority

### DIFF
--- a/ds_caselaw_editor_ui/templates/includes/judgments_list_item.html
+++ b/ds_caselaw_editor_ui/templates/includes/judgments_list_item.html
@@ -1,3 +1,4 @@
+{% load waffle_tags %}
 {% load static i18n %}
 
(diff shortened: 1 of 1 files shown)


Terry-Price:
This is so much cleaner especially if only use it on Jira.

Terry-Price (approved):
this is so much cleaner – especially if they use it only on Jira.

jacksonj04 (commented):
Just to double check you've got the logic on this the right way round - we're defaulting to the priority being removed, unless we specifically flag for an editor to see it?

PR #716, opened 2023-02-27
Redesign URL structure to be cleaner and more consistent

jacksonj04:
This updates the URL structure of the EUI to be cleaner and more consistent with the end goal of the PUI. In particular, it gets rid of query parameters and is more explicit in the URLs about what to expect.

In particular, we move from URLs which look like `/edit?judgment_uri=ewhc/kb/2023/376` to ones which look like `/ewhc/kb/2023/376/edit`.

**This intentionally doesn't cover all of the verbs**, since things like `delete` and `unlock` don't follow a RESTful pattern.


dragon-dxw:
Big and scary but all looks good. I think there's some work in our future to better define what a judgment_uri path is but we don't need it yet.

judgments/urls.py
```diff
@@ -6,26 +6,33 @@
...
     path("delete", delete, name="delete"),
     path("unlock", unlock, name="unlock"),
     path("assign", assign_judgment_button, name="assign"),
     path("prioritise", prioritise_judgment_button, name="prioritise"),
     path("hold", hold_judgment_button, name="hold"),
-    # redirect to signed asset URLs
-    path("signed-asset/<path:key>", redirect_to_signed_asset, name="signed-asset"),
-    path("", index, name="home"),
+    # Redirects for legacy judgment URIs
+    path("edit", edit_view_redirect),
```
anthonyhashemi, line 30: So basically we have new more RESTful urls for the judgement endpoints but  the legacy url endpoints still exist but are now redirected to the same views that are used for the new endpoints (which have the same logic as before anyway).

jacksonj04: That's the plan - the only changes to the actual views are where we're now using the named parameters rather than `request.GET`. The only exception is the unlock view, where we now additionally initialise a proper `Judgment` object to get a cleaned `uri` parameter (otherwise there's a stray slash introduced to the final redirect, and it all works, but it looks... ugh).

judgments/urls.py
```diff
@@ -6,26 +6,33 @@
...
-    path("xml", detail_xml, name="detail_xml"),
+    # Home
+    path("", index, name="home"),
+    # Search
     path("results", results, name="results"),
-    # buttons that do a thing and redirect
+    # redirect to signed asset URLs
+    path("signed-asset/<path:key>", redirect_to_signed_asset, name="signed-asset"),
+    # Legacy judgment verbs
     path("delete", delete, name="delete"),
```
anthonyhashemi, line 24: I think having the ability to make a HTTP DELETE request:
1. on a specific resource (id) `DELETE /id/`
2. on many from the collection by specifying some filtering with `DELETE /`

are both RESTful

judgments/urls.py
```diff
@@ -6,26 +6,33 @@
...
     path("hold", hold_judgment_button, name="hold"),
-    # redirect to signed asset URLs
-    path("signed-asset/<path:key>", redirect_to_signed_asset, name="signed-asset"),
-    path("", index, name="home"),
+    # Redirects for legacy judgment URIs
+    path("edit", edit_view_redirect),
+    path("detail", html_view_redirect),
+    path("xml", xml_view_redirect),
+    # Different views on judgments
+    path("<path:judgment_uri>/edit", EditJudgmentView.as_view(), name="edit-judgment"),
```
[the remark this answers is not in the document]
jacksonj04: No, because this path is the one which displays the edit form as well as the one which receives the `POST`!

In an ideal world yes, I'd love to retool things so that the form does a `PATCH` to the resource root (well, really a `POST` with a method specified because HTML forms can't do any verbs except that and `GET`), but smol steps!

judgments/urls.py
jacksonj04: Somewhat related to [`PUT`ting updates to the resource](https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/716#discussion_r1120480097), we need to do some clever magic with form methods so that we can just direct all these things to the resource rather than to verb-based URIs!

judgments/urls.py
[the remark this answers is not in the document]
anthonyhashemi: ah I see - I need to do some reading on html forms!

judgments/urls.py
anthonyhashemi: yeah, I will need to do some reading up on html forms. But let me know if you have some time to jump on a call later today to discuss this in a bit more detail

anthonyhashemi (approved):

PR #717, opened 2023-02-27
Fix/reduce home banner

Terry-Price:
## Changes in this PR:
Changed the Unpublished judgments (home) screen header so it is much shallower to reduce unnecessary scrolling.
Also reduce the space between the slim header and the page content to also reduce page scroll further.

https://trello.com/c/gG80HYsN/496-da-%E2%9C%8F%EF%B8%8F-eui-header-design
### Before
Screenshot 2023-02-27 at 14 38 12

### After
Screenshot 2023-02-27 at 14 37 57


Donna-H (approved):
this looks fine

PR #729, opened 2023-02-28
Add 'report a problem' link

jacksonj04:
This link goes to a Google Form which tries to capture some (very loosely) structured information on what has gone wrong. On submission, this is converted to a Trello card and flagged in Slack.

This is behind the `report_problem_link` feature flag.


dragon-dxw (approved):

ds_caselaw_editor_ui/templates/includes/report_a_problem_link.html
```diff
@@ -0,0 +1 @@
+<p><a target="_blank" draggable="false" rel="noreferrer noopener" href="https://docs.google.com/forms/d/e/1FAIpQLSfpRGyDYMXVVDwgkYyD4fsHTwBN0XVxoZTMHePDB_hmmrw2sA/viewform?usp=pp_url&entry.2130513657={{ request.get_full_path|urlencode }}">Report a problem with this page</a></p>
```
dragon-dxw, line 1: untranslated string, but I suspect we'll need a different form anyway if it's in Welsh.

jacksonj04:
Parking this back in draft because we're working on a new unified feedback/problem form and pipeline.

PR #748, opened 2023-03-07
Fixed LocalStack setup script

jacksonj04:
This script wasn't working. Now it is.
Maybe it's changing versions of localstack to v1.


script/setup-localstack.sh
```diff
@@ -3,7 +3,7 @@ source .env
 awslocal sns create-topic --name caselaw-stg-judgment-updated
 awslocal s3api create-bucket --bucket private-asset-bucket
 awslocal s3api create-bucket --bucket public-asset-bucket
-awslocal --endpoint-url=http://localhost:4566 sqs create-queue --queue-name dummy-queue  --region us-east-1 --output table
```
dragon-dxw, line 6: the endpoint URLs should be consistent, so we consistently removed them.

dragon-dxw (approved):

PR #750, opened 2023-03-08
Fix mixed display styles for the breadcrumb bar

jacksonj04:
The breadcrumb bar had varying vertical padding in different places after removal of the logo. This is now fixed and the displaying of the breadcrumb bar simplified.


ds_caselaw_editor_ui/templates/layouts/base.html
```diff
@@ -38,8 +38,12 @@
 
     <a id="skip-to-main-content" href="#main-content">{% translate "skiplink" %}</a>
 
-    {% block header %}
-    {% endblock header %}
+    <header class="page-header page-header-slim">
+      {% include 'includes/breadcrumbs.html' with current=context.page_title title=context.page_title link=request.get_full_path %}
```
anthonyhashemi, line 42: so basically we just define the breadcrumbs in the base.html which is imported / extended in each other template anyway as opposed to defining multiple times which led to inconsistency.

ds_caselaw_editor_ui/templates/layouts/base.html
```diff
@@ -38,8 +38,12 @@
...
+    </header>
+
+    {% block precontent %}
```
anthonyhashemi, line 45: this precontent stuff is new, whats this about?

ds_caselaw_editor_ui/templates/layouts/judgment.html
```diff
@@ -1,14 +1,6 @@
...
 
-{% block header %}
-
-  <header class="page-header page-header-slim">
-
-    {% include 'includes/breadcrumbs.html' with current=context.page_title title=context.page_title link=request.get_full_path %}
-
-  </header>
-
+{% block precontent %}
```
anthonyhashemi, line 4: likewise here

ds_caselaw_editor_ui/templates/layouts/base.html
anthonyhashemi, line 42: is there a reason we dont want the logo that was in some of the breadcrumbs?

ds_caselaw_editor_ui/templates/layouts/base.html
jacksonj04: This is to hold the judgment toolbar which previously lived in the `{% block header %}`, which should come before the `<main>` in the HTML to keep our semantics clean.

ds_caselaw_editor_ui/templates/layouts/base.html
jacksonj04: Yep - we had various flavours scattered about, some were in `header` blocks, some are in the page content proper, some were using the `-slim` class and some weren't... so now we have the one place for all header stuff.

jacksonj04: It boils down to TNA's branding guidelines - if we want to include the logo it _has_ to be quite big, but we're assuming that everyone using the EUI knows where they work so there's no real benefit to it and it just takes up a load of space.

anthonyhashemi: ah I see

ds_caselaw_editor_ui/templates/layouts/base.html
anthonyhashemi: cool 👍

anthonyhashemi (approved):

PR #752, opened 2023-03-08
Create new labs view

jacksonj04:
For features currently undergoing testing, this allows editors to self-serve and enable or disable flags for their current session.

## Screenshots

localhost_3000_labs

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_header.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_header.scss
@@ -2,7 +2,12 @@
   background-color: $color__almost-black;
   padding: 0 0 calc($spacer__unit * 1.25);
 
+  display: flex;
+  flex-direction: row;
+
   &__breadcrumb {
+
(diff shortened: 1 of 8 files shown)


ds_caselaw_editor_ui/sass/includes/_header.scss
```diff
@@ -2,8 +2,13 @@
   background-color: $color__almost-black;
   padding: 0 0 calc($spacer__unit * 1.25);
 
+  display: flex;
```
dragon-dxw, line 5: what lead to the need for the changes to the header?

dragon-dxw (approved):

ds_caselaw_editor_ui/sass/includes/_header.scss
jacksonj04: It's the easiest way of saying "all this stuff on the left takes as much space as it can, and then this other bit goes off to the right".

PR #751, opened 2023-03-08
Add metadata box to judgment full-text view

jacksonj04:
nb: This knowingly has low test coverage; the entire update judgment flow is being refactored to be more testable and robust.

PR #758, opened 2023-03-09
Revert unwanted changes to judgments_list_item.html

jacksonj04:
This fixes a regression where list items in the home page were trying to use legacy URL names.

### DIFF
--- a/ds_caselaw_editor_ui/templates/includes/judgments_list_item.html
+++ b/ds_caselaw_editor_ui/templates/includes/judgments_list_item.html
@@ -1,3 +1,4 @@
+{% load waffle_tags %}
 {% load static i18n %}
 
 <li class="judgments-list__judgment">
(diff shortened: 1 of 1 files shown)


anthonyhashemi (approved):

ds_caselaw_editor_ui/templates/includes/judgments_list_item.html
```diff
@@ -1,3 +1,4 @@
+{% load waffle_tags %}
```
anthonyhashemi, line 1: Reverts the unwanted changes from: https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/751/commits/19ce3a724163daed7c35a808c46a3d60a6510aee#diff-1892d6ca0fdcad0ea387650aa14bad1e05129afdca201332d28146646c047078.

Explanation from Nick from slack:

"
the root of the problem is that the (current) edit-judgment route name was changed back to its (old) edit route name, but that problem only appears if judgments are assigned. Which none of the ones on local test or staging are.
"

PR #759, opened 2023-03-09
Refactor "publish a judgment" workflow

jacksonj04:
At the moment, the process of (un)publishing a judgment is handled by `edit_judgment.py` making API calls.

Since we now have the `Judgment` object, we can implement new `publish()` and `unpublish()` functions which do the same thing in a more contained and testable way.

### DIFF
--- a/judgments/models/judgments.py
+++ b/judgments/models/judgments.py
@@ -58,6 +58,10 @@ class Judgment:
         return self.api_client.get_published(self.uri)
 
     @cached_property
+    def is_held(self) -> bool:
+        return self.api_client.get_property(self.uri, "editor-hold")
+
+    @cached_property
     def is_sensitive(self) -> bool:
         return self.api_client.get_sensitive(self.uri)
 
--- a/judgments/models/judgments.py
+++ b/judgments/models/judgments.py
@@ -1,0 +1,133 @@
+import datetime
+from functools import cached_property
+from typing import Optional
+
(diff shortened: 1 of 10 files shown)


judgments/tests/tests.py
```diff
@@ -5,7 +5,8 @@
 from django.urls import reverse
 from lxml import etree
 
-from judgments.models import Judgment, SearchResult, SearchResultMeta
+from judgments.models import SearchResult, SearchResultMeta
+from judgments.models.judgments import Judgment
```
dragon-dxw, line 9: Would it be nicer to import Judgment into judgments.models, so all models are in one place? (Would mean we'd lose much of the second commit)

judgments/models/judgments.py
```diff
@@ -57,6 +57,10 @@ def judgment_date_as_date(self) -> datetime.date:
     def is_published(self) -> bool:
         return self.api_client.get_published(self.uri)
 
+    @cached_property
```
dragon-dxw, line 71: wary of the caching on this, since it'll be mostly used in the editor; how does the cache get invalidated?

dragon-dxw: (it occurs somewhere else too)

dragon-dxw:
We had a chat about the scope of the caching -- it's literally that Python object, so won't be a problem

dragon-dxw (approved):

PR #762, opened 2023-03-09
Autofill court codes

dragon-dxw:
Uses a datalist to show a drop-down list of options; we include codename and ID to facilitate multiple types of search.

Changes width of court and neutral citation fields to be a little wider by default.

Should fall back gracefully if the datalist is not supported.

Strange bug on my system where "A-COURT" comes up as an option -- it's not in the HTML, so I assume it's my browser.

Do we want a feature flag on this?

image

### DIFF
--- a/judgments/views/full_text.py
+++ b/judgments/views/full_text.py
@@ -1,5 +1,6 @@
 from urllib.parse import urlencode
 
+import ds_caselaw_utils as caselawutils
 import waffle
 from caselawclient.Client import MarklogicResourceNotFoundError
 from django.http import Http404, HttpResponse, HttpResponseRedirect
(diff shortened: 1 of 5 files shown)


timcowlishaw (approved):
:shipit:

jacksonj04 (changes requested):
I'm not bothered for this being behind a flag, it's a good general-purpose improvement with no huge functional changes.

Could we also add it to the metadata box on the HTML/PDF view?

jacksonj04 (approved):

judgments/views/full_text.py
```diff
@@ -16,7 +17,11 @@ def html_view(request, judgment_uri):
 
     try:
         judgment = Judgment(judgment_uri)
-        context = {"judgment_uri": judgment_uri, "judgment": judgment}
+        context = {
```
anthonyhashemi, line 20: I think it's worth adding a test for this changed logic

PR #771, opened 2023-03-13
Simplify secondary button class

Terry-Price:
## Changes in this PR:
Decided to reduce button styles on the EUI. So that we had one class for the secondary button that can be used for all secondary buttons.
[This was](https://trello.com/c/upm0qF6y/603-eui-fix-reduce-sass-styles-of-secondary-button-to-just-one)

### Before
This shows multiple classes for the button
Screenshot 2023-03-13 at 15 49 40

### After
This shows a single class for the button.
Screenshot 2023-03-13 at 15 50 03

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_buttons.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_buttons.scss
@@ -1,0 +1,9 @@
+.button-secondary {
+        @include call-to-action-button;
+        background-color: transparent;
+        color: $color__cta-background !important;
+        border: 2px solid $color__cta-background;
+    &:visited  {
+      color: #ffffff;
+    }
+}
--- a/ds_caselaw_editor_ui/sass/includes/_how_can_this_service_be_improved.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_how_can_this_service_be_improved.scss
@@ -12,11 +12,4 @@
         font-size: 1.3rem;
       }
     }
-
-    &__cta-button {
-      @include call-to-action-button-secondary;
-      &:visited  {
-        color: #ffffff;
-      }
-    }
   }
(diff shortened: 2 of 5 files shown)


ds_caselaw_editor_ui/sass/includes/_buttons.scss
```diff
@@ -0,0 +1,9 @@
+.button-secondary {
+        @include call-to-action-button;
+        background-color: transparent;
+        color: $color__cta-background !important;
```
dragon-dxw, line 4: we're still referring to the cta button in this colour definition -- is that intentional?

timcowlishaw (approved):
This looks good to me - just two outstanding questions / requests:

- Are we confident that this is now being used everywhere on the editor site there's a 'secondary' button, or might there be more duplicate secondary button styles we haven't discovered yet hanging around?
- Can we get a PR (or at least a pending trello card) to do the same refactor on public? I think we've got a discussion pending that should happen during this refactoring work about how best to share CSS between the two applications, but in the meantime, while we're maintaining two seperate codebases, it'd be good to make sure they don't diverge too much.

Neither of these should block merging, as long as they're captured as pending bits of work to do.

ds_caselaw_editor_ui/sass/includes/_buttons.scss
timcowlishaw: aha i'll check with Terry if it's intentional or not, but i can see the logic, if this button is a variation on the other, we reuse the styles and just describe how it's different. (then if the overall colour pallete changed, we would still have buttons that are coloured consistently with each other, for instance)

Terry-Price:
Checked all placed the version of button has been used and happy that it's covered in all places.

PR #775, opened 2023-03-14
Judgment status indicator

jacksonj04:
Add a status indicator for a judgment to the sidebar, showing one of "in progress", "published" or "on hold".

The tag formatting is based on the GOV.UK Design System.

### Published

A judgment is "published" if its published flag is set _regardless of if it is on hold or not_.

### On hold

A judgment is "on hold" if its hold flag is set _and_ it is not published.

### In progress

A judgment is "in progress" if it has neither published or hold flags set.

### DIFF
--- a/judgments/models/judgments.py
+++ b/judgments/models/judgments.py
@@ -17,6 +17,10 @@ from judgments.utils.aws import (
     uri_for_s3,
 )
 
+JUDGMENT_STATUS_HOLD = "On hold"
+JUDGMENT_STATUS_PUBLISHED = "Published"
+JUDGMENT_STATUS_IN_PROGRESS = "In progress"
+
 
 class CannotPublishUnpublishableJudgment(Exception):
     pass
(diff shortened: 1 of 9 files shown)


judgments/models/judgments.py
```diff
@@ -154,6 +158,16 @@ def is_publishable(self) -> bool:
 
         return True
 
+    @property
+    def status(self) -> str:
```
dragon-dxw, line 162: status_message?

judgments/templatetags/status_tag_css.py
```diff
@@ -0,0 +1,22 @@
...
+    JUDGMENT_STATUS_PUBLISHED,
+)
+
+register = template.Library()
+
+
+@register.filter
+@stringfilter
+def status_tag_colour(status):
+    if status == JUDGMENT_STATUS_IN_PROGRESS:
```
dragon-dxw, line 16: This starts to feel like we should have a dictionary of statuses

dragon-dxw (approved):

judgments/templatetags/status_tag_css.py
jacksonj04: I debated it, so we really strictly separate the presentation strings from the status string.

PR #780, opened 2023-03-15
Use TDR reference in full-text metadata panel, not URI

jacksonj04:
This now matches the label.


dragon-dxw (approved):
Assuming the backend already does this :)

PR #786, opened 2023-03-16
Create styleguide page

timcowlishaw:
## Changes in this PR:

Create a styleguide and populate it with our button styles as a placeholder example (the CTA button is, ironically, unstyled, but that's because it was abstracted out as a reusable component on another branch that's pending merge - it'll magically become styled when that's done!)

This is going to offer front end devs a place to develop reusable components in isolation, as well as acting as a pallete of reusable components, breaking the dependency of front end dev on backend requirements, encouraging us to write reusable code, and documenting our work (at least in theory!)

https://trello.com/c/o7AZSoyw/609-eui-start-creating-a-style-guide-page

## Picture (for posterity)
Screenshot 2023-03-17 at 11 35 49

### DIFF
--- a/judgments/urls.py
+++ b/judgments/urls.py
@@ -26,6 +26,7 @@ from .views.judgment_publish import (
 from .views.labs import Labs
 from .views.results import results
 from .views.signed_asset import redirect_to_signed_asset
+from .views.style_guide import StyleGuide
 from .views.unlock import unlock
 
 urlpatterns = [
@@ -49,6 +50,7 @@ urlpatterns = [
     path("xml", xml_view_redirect),
     # Labs
     path("labs", Labs.as_view(), name="labs"),
+    path("style_guide", StyleGuide.as_view(), name="style_guide"),
     # Different views on judgments
     path("<path:judgment_uri>/edit", EditJudgmentView.as_view(), name="edit-judgment"),
     path(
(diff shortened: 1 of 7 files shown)


judgments/urls.py
```diff
@@ -26,6 +26,7 @@
 from .views.labs import Labs
 from .views.results import results
 from .views.signed_asset import redirect_to_signed_asset
+from .views.styleguide import Styleguide
```
dragon-dxw, line 29: It bugs me that Styleguide is one word, but you're 100% consistent so I guess that's correct :D

ds_caselaw_editor_ui/templates/pages/styleguide.html
```diff
@@ -0,0 +1,10 @@
+{% extends "layouts/base.html" %}
+{% load i18n %}
+{% block content %}
+  <div class="styleguide">
```
dragon-dxw, line 4: hot take: we shouldn't have any custom CSS for the styleguide; it should use a generic page style which can be easily used by other pages

dragon-dxw (approved):
Two spicy comments, but if you're happy, I'm happy. :)
I'd also like pictures if possible, so people can see how this looked back in the day.

judgments/urls.py
timcowlishaw: that also bugs me, tbh. Fixed!

ds_caselaw_editor_ui/templates/pages/styleguide.html
timcowlishaw: Very good hot take, and i think that can be our first css refactor - the reason i added it was because we didn't have a default page layout style, which we should do! I'll merge as is, but put that at the top of the refactoring list :-)

PR #803, opened 2023-03-24
Style guide/tabs

Terry-Price:
https://trello.com/c/yXaeElqW/647-eui-style-tab-switcher-for-html-pdf-view
### Before
tabs-placeholder-design

### After
tabs-component

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_tabs.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_tabs.scss
@@ -4,22 +4,29 @@
 
   &__item {
     display: inline-block;
-    margin: auto .25rem;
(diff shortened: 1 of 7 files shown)


timcowlishaw (approved):
Approving, but i think Nick should have the final word :-)

timcowlishaw:
Just wanted to note, while there's JS in here, this will work fine without it too :-)

timcowlishaw:
I would like at some point to extend this to cope with more than 2 tabs, but that can wait 'til needed. It ought to be pretty simple, just a matter of refactoring the template a little

jacksonj04 (approved):
Looks nice, very sharp!

PR #808, opened 2023-03-24
Put a judgment on hold

Donna-H:
## Changes in this PR:
Create a Hold flow for editors to put documents on hold. This should be behind a feature flag for now,
Button states of the publish and ‘on hold’ buttons to ensure that the right actions are available in each state, and invalid ones are shown as disabled.
_**ALERT**_ - on hold button needs backend fix - onhold page has status not updated due to error in hold button. Therefore status will say on hold when button is fixed. Also as hold button is not putting the document on hold the Unhold button will not enable.  See for fix this Trello card here ; https://trello.com/c/Djy2fwfL/696-server-side-regression-putting-judgments-on-and-off-hold

https://trello.com/c/bvTonm8L/654-eui-as-an-editor-i-want-to-put-a-judgment-on-hold-in-order-to-signal-to-other-editors-and-the-submitter-that-it-cant-be-publishe
### Before
before-hold

### After
If published cannot go on hold 
Judgment published: Put on hold [Disabled], Publish [Disabled] Unpublish [Enabled]
published-no-hold

Judgment unpublished: Put on hold [Enabled], Publish [Enabled]
Put on hold
put-on-hold-page

Onhold - **Alert** see Trello above as hold button is not putting the document on hold so Unhold button will not enable
Judgment on hold: Unhold [Enabled], Publish [Disabled]

onhold-page

Unhold **Alert** see Trello above as hold button is not putting the document on hold so Unhold button will not enable
unhold-request

Unheld

unheld-page


ds_caselaw_editor_ui/templates/judgment/hold-success.html
```diff
@@ -0,0 +1,16 @@
...
+{% load i18n %}
+
+
+{% block judgment_header %}
+  {% include "includes/judgment_metadata_panel.html" with judgment=context.judgment %}
+{% endblock %}
+
+
+{% block judgment_content %}
+  <h2>This document has been put on hold</h2>
```
timcowlishaw, line 12: If you could pop this string in the translations, that'd be ace, I've added some updated instructions to README.md which should help!

ds_caselaw_editor_ui/templates/judgment/hold.html
```diff
@@ -0,0 +1,31 @@
+{% extends "layouts/judgment_with_sidebar.html" %}
+{% load waffle_tags %}
+{% load i18n %}
+
+{% block judgment_header %}
+  {% include "includes/judgment_metadata_panel.html" with judgment=context.judgment %}
+{% endblock %}
+
+{% block judgment_content %}
+  <h2>Put this document on hold</h2>
```
timcowlishaw, line 10: Ditto, it'd be great if this was in translations

ds_caselaw_editor_ui/templates/judgment/unhold-success.html
```diff
@@ -0,0 +1,13 @@
...
+{% load waffle_tags %}
+{% block judgment_content %}
+  {% load i18n %}
+  <h2>
+    {% translate 'judgment.hold.unhold_success_title' %}
+    <br>
+    {{ context.judgment.name }}
+  </h2>
+
+  <p>This document is no longer on hold.</p>
```
timcowlishaw, line 11: More stray translations :-)

judgments/views/judgment_hold.py
```diff
@@ -0,0 +1,161 @@
...
+        judgment = Judgment(judgment_uri)
+        judgment.hold()
+        invalidate_caches(judgment.uri)
+        messages.success(request, gettext("judgment.hold.hold_success_flash_message"))
+        return HttpResponseRedirect(
+            reverse("hold-judgment-success", kwargs={"judgment_uri": judgment.uri})
+        )
+    except MarklogicResourceNotFoundError:
+        return HttpResponseBadRequest(
+            escape(f"Judgment {judgment_uri} could not be found.")
```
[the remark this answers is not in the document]
timcowlishaw: @jacksonj04 any opnions on the automatic alert here? i think it's fairly low risk as this is editors only but will defer to you

timcowlishaw (changes requested):
Looks good, Donna! A couple of strings in templates that should go in translations, and a question for @jacksonj04 to answer, but otherwise i'm happy

ds_caselaw_editor_ui/sass/includes/_judgment_component.scss
```diff
@@ -94,6 +94,23 @@
     padding-bottom: 10px;
   }
 }
+.email-hold-publish {
+  border: 2px #ffb952 solid;
+  display: inline-block;
+  padding: 0px 10px 0px 0;
```
timcowlishaw, line 100: Little note for the future, but not a blocker for this PR - it'd be good to use `calc` to make padding and margin relative to the `$spacer__unit` variable in future where possible

judgments/views/judgment_hold.py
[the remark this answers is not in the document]
jacksonj04: Actually scratch that, it [_did_ flag in the publish flow](https://github.com/nationalarchives/ds-caselaw-editor-ui/security/code-scanning/22), I suspect because it's not figured out what `django.utis.html.escape` is doing here.

The more 'proper' way to solve this is to escape the string we're using in the first place, so:

```suggestion
            "Judgment {judgment_uri} could not be found.".format(judgment_uri=escape(judgment_uri)
```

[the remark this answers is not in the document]
Donna-H: @jacksonj04 okay great ! Will update it with this 👍🏾

ds_caselaw_editor_ui/templates/judgment/hold-success.html
Donna-H: okay will do thank you !

jacksonj04 (changes requested):
I think this could do with some tests writing just to make sure it's behaving as it should. If you need a hand writing them please hit me up!

ds_caselaw_editor_ui/sass/includes/_judgment_component.scss
```diff
@@ -94,6 +94,23 @@
     padding-bottom: 10px;
   }
 }
+.email-hold-publish {
```
jacksonj04, line 97: This seems like a very specific style; is this actually exclusive to the hold and publish pages?

ds_caselaw_editor_ui/templates/judgment/unhold.html
```diff
@@ -0,0 +1,31 @@
...
+      />
+      <input
+        type="hidden"
+        name="unhold"
+        value="false"
+      />
+      <input
+        class="button-cta"
+        type="submit"
+        value="{% translate 'judgment.unhold.unhold_button' %}"
```
jacksonj04, line 26: Super nitpicky, but could we tweak this to:
```suggestion
        value="{% translate 'judgment.hold.unhold_button' %}"
```
so it's more consistent with the pattern in publish?

ds_caselaw_editor_ui/templates/judgment/unhold-success.html
```diff
@@ -0,0 +1,13 @@
...
+{% load waffle_tags %}
+{% block judgment_content %}
+  {% load i18n %}
+  <h2>
+    {% translate 'judgment.hold.unhold_success_title' %}
+    <br>
+    {{ context.judgment.name }}
+  </h2>
+
+  <p>{% translate "judgment.unhold.success" %}</p>
```
jacksonj04, line 11:
```suggestion
  <p>{% translate "judgment.hold.unhold_success_message" %}</p>
```
Again super nitpicky, but I think this better conveys the intent of this string.

judgments/models/judgments.py
```diff
@@ -74,7 +74,7 @@ def is_published(self) -> bool:
 
     @cached_property
     def is_held(self) -> bool:
-        return self.api_client.get_property(self.uri, "editor-hold")
+        return self.api_client.get_property(self.uri, "editor-hold") == "true"
```
jacksonj04, line 77: That's interesting; if `get_property` isn't returning a bool already how is mypy not tripping up? We must not be checking into the dependency. One to fix later, not relevant to this PR.

judgments/models/judgments.py
```diff
@@ -188,3 +188,9 @@ def unpublish(self):
             status="not published",
             enrich=False,
         )
+
+    def hold(self):
+        self.api_client.set_property(self.uri, "editor-hold", "true")
```
jacksonj04, line 193: The fact this isn't a boolean makes me sad, but not a problem for this PR!

Do we have any business logic we want to apply here, eg "you cannot hold a published judgment"?

judgments/urls.py
```diff
@@ -73,6 +83,26 @@
         UnpublishJudgmentSuccessView.as_view(),
         name="unpublish-judgment-success",
     ),
+    path(
+        "<path:judgment_uri>/hold",
+        HoldJudgmentView.as_view(),
+        name="hold-judgment",
+    ),
+    path(
+        "<path:judgment_uri>/onhold",
```
jacksonj04, line 92:
```suggestion
        "<path:judgment_uri>/held",
```

jacksonj04 (approved):
Couple of tiny suggestions, but nothing blocking I can see!

PR #811, opened 2023-03-27
Chore/switch to djlint

timcowlishaw:
## Changes in this PR:

- Introduce djlint to normalise formatting and lint html templates on push
- Reformat all our templates in bulk to accomodate it
- Fix some slightly brittle tests that cared about whitespace when they shouldn't


dragon-dxw:
[Swap curlylint for djlint](https://trello.com/c/JK7aBPau/672-swap-curlylint-for-djlint)

PR #818, opened 2023-03-28
Update translations file

jacksonj04:
The translations file has become out of sync with the ordering gettext generates; manually regenerate it to bring this into alignment and reduce the size of unrelated diffs.


timcowlishaw (commented):
aha, thanks for this @jacksonj04  - just one quick question - is it worth doing this with --no-location and just blitzing them now?

PR #819, opened 2023-03-28
Style guide/validation messages

Terry-Price:
## Changes in this PR:
Created a set components for validation messaging for success and failure that can be used for page load events and for individual elements such as forms etc.
https://trello.com/c/2mpSy3GD/634-implement-design-for-success-or-other-message-banners-so-it-is-consistent-across-the-site
### Before
Screenshot 2023-03-28 at 13 52 37

### After
Screenshot 2023-03-28 at 13 49 29
Screenshot 2023-03-28 at 13 50 59


jacksonj04 (commented):
I'm not sure "validation" is the right word here; perhaps "notification" is more aligned with what they actually do (this is what [GDS use](https://design-system.service.gov.uk/components/notification-banner/)). They're often used to notify the user about the outcome of validation, but we also use them (especially the page-level one) to relay that an action has been successful.

timcowlishaw:
> perhaps "notification" is more aligned with what they actually do (this is what [GDS use](https://design-system.service.gov.uk/components/notification-banner/))

Aha, i agree with this actually - the elements don't 'validate' themselves but _notify_ the results (and even better to be consistent with GDS language). I guess the other term people use is 'flash message' but i suspect that's a little too esoteric. Definitely support renaming to 'notification' unless there's any strong arguments against!

timcowlishaw:
Oh btw, when this is approved @Terry-Price I'll quickly go through and make sure it's used everywhere it needs to be :-)

timcowlishaw (approved):

PR #825, opened 2023-03-29
Remove doubled 'context' in template parameters.

timcowlishaw:
## Changes in this PR:

As well as being confusing this was causing problems for template re-use, where an include was used in two places - one with the extra 'context' variable on the context, and another without. This removes the extra 'context' everywhere it's used for consistency - further refactoring to come to tidy up the views and remove duplication.


dragon-dxw (approved):
Gone top down but eyes were glazing over, the word context has lost all meaning. Recommend someone else scans from bottom up.

jacksonj04 (approved):
Is "context" even a real word any more?

🚢

PR #829, opened 2023-03-30
Use new flash notification messages site-wide

timcowlishaw:
This commit uses terry's new notification component for flash messages, and removes some stray additional styling from before.

## Screenshots of UI changes:
Screenshot 2023-03-30 at 12 56 38


timcowlishaw:
New improved placement: 
imagen

Terry-Price:
looks good to me.

PR #832, opened 2023-03-30
Restrict list of users in the assignment drop-down menu

jacksonj04:
At the moment, the 'assign users' dropdown shows every user in the system. This PR adds the option of providing an `EDITORS_GROUP_ID` env variable which, if present, will restrict display to only active users in that group.


judgments/utils/__init__.py
```diff
@@ -141,12 +143,20 @@ def extract_version(version_string: str) -> int:
...
-            "print_name": u.get_full_name() or u.get_username(),
-        }
-        for u in users
-    ]
+def editors_dict():
+    if settings.EDITORS_GROUP_ID:
+        editors_group = Group.objects.get(id=settings.EDITORS_GROUP_ID)
+        editors = editors_group.user_set.filter(is_active=True)
+    else:
+        editors = User.objects.filter(is_active=True)
```
dragon-dxw, line 151: Since we've got two paths with .filter(is_active=True), we're only testing one of them

dragon-dxw (approved):

PR #835, opened 2023-03-31
Use new tabs site wide

Donna-H:
## Changes in this PR:
Add the new tab from style guide so it is available site wide -put behind the embedded pdf flag 
https://trello.com/c/AIJddSdz/740-eui-style-tab-switcher-for-html-pdf-view
### Before
button-tab

### After
[tab-toggle.webm](https://user-images.githubusercontent.com/75584408/229161384-0b9bbab6-6ca3-4020-95c5-35de1a2168bb.webm)

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_judgment_view_controls.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_judgment_view_controls.scss
@@ -1,0 +1,7 @@
+.judgment-view-controls{
+
+  @media (min-width: 1400px) {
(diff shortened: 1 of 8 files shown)


Terry-Price (approved):
Looks Good Donna.
I did notice that the PDF does not load on your browser after looking at the video, which I remembered to watch for the first ever time.

Donna-H:
@Terry-Price yes this message should always appear for users where their browser cannot display the pdf with an option to download it instead.

PR #841, opened 2023-04-04
Add assignment dropdown to judgment sidebar

timcowlishaw:
## Changes in this PR:

This allows editors to reassign judgments on the publish, hold and unhold pages.

It relies on javascript for the behaviour given in the axure prototypes - without JS it will gracefully degrade to a standard form with a submit button.

I've also added test coverage for the assignment button handler logic.

https://trello.com/c/T9J35JZF/712-refinement-move-assignment-to-sidebar

## Screenshots of UI changes:

Screenshot 2023-04-04 at 08 08 42
Screenshot 2023-04-04 at 08 08 31

### DIFF
--- a/ds_caselaw_editor_ui/static/js/src/components/judgmentSidebar.js
+++ b/ds_caselaw_editor_ui/static/js/src/components/judgmentSidebar.js
@@ -1,0 +1,48 @@
+import $ from "jquery";
+
+const JudgmentSidebar = {
+  init: function () {
+    const elements = $(".judgment-sidebar");
+    elements.each(function (ix, element) {
+      JudgmentSidebar.initOne(element);
+    });
+  },
+
+  initOne: function (element) {
+    JudgmentSidebar.initAssignForm(element);
+  },
+
+  initAssignForm: function (element) {
+    const form = $(element).find(".judgment-sidebar__judgment-assign-form");
+    const select = form.find("select");
+    const submit = form.find("input[type='submit']");
+    submit.hide();
+    select.on("change", function (event) {
+      form.find(".judgment-sidebar__context-notification").remove();
+      form.trigger("submit");
+    });
+    form.on("submit", function (event) {
+      event.preventDefault();
+      const form = $(this);
+      const action = form.attr("action");
+      const loading = $(
+        "<span class='loading-indicator' role='progressbar' aria-valuetext='Loading' aria-busy='true' aria-live='polite'></span>"
+      );
+      form.append(loading);
+      $.ajax({
+        type: "POST",
+        url: action,
+        data: form.serialize(),
+        success: function (data) {
+          const successMessage = $(
+            "<div class='context-notification--success judgment-sidebar__context-notification' aria-busy='false' />"
+          );
+          successMessage.text(data["message"]);
+          loading.replaceWith(successMessage);
+        },
+      });
+    });
+  },
+};
+
+export default JudgmentSidebar;
--- a/ds_caselaw_editor_ui/templates/includes/judgment_sidebar.html
+++ b/ds_caselaw_editor_ui/templates/includes/judgment_sidebar.html
@@ -46,21 +46,32 @@
     </li>
   </ul>
 </div>
-{% if feature_flag_assign_in_publish_sidebar %}
+{% if feature_flag_publish_flow %}
   <div class="judgment-sidebar__block">
     <h4>{% translate "judgments.sidebar.assigned_to" %}</h4>
-    <select class="judgment-component__assigned_to-input"
-            name="assigned_to"
-            id="assigned_to">
-      <option value="" {% if judgment.assigned_to == "" %}selected{% endif %}>
-        {% translate "judgments.noone" %}
-      </option>
-      {% for editor in editors %}
-        <option value="{{ editor.name }}"
-                {% if judgment.assigned_to == editor.name %} selected{% endif %}>
-          {{ editor.print_name }}
+    <form action="{% url "assign" %}"
+          method="post"
+          class="judgment-sidebar__judgment-assign-form">
+      {% csrf_token %}
+      <input type="hidden" name="judgment_uri" value="{{ judgment.uri }}" />
+      <select class="judgment-sidebar__assignment-input"
+              name="assigned_to"
+              id="assigned_to">
+        <option value="" {% if judgment.assigned_to == "" %}selected{% endif %}>
+          {% translate "judgments.noone" %}
         </option>
-      {% endfor %}
-    </select>
+        {% for editor in editors %}
+          <option value="{{ editor.name }}"
+                  {% if judgment.assigned_to == editor.name %} selected{% endif %}>
+            {{ editor.print_name }}
+          </option>
+        {% endfor %}
+      </select>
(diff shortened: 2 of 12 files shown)


ds_caselaw_editor_ui/templates/includes/judgment_sidebar.html
```diff
@@ -49,18 +49,25 @@ <h4>{% translate "judgments.sidebar.tools" %}</h4>
 {% if feature_flag_assign_in_publish_sidebar %}
```
jacksonj04, line 49: This is the wrong flag for this feature.

ds_caselaw_editor_ui/templates/includes/judgment_sidebar.html
```diff
@@ -49,18 +49,25 @@ <h4>{% translate "judgments.sidebar.tools" %}</h4>
...
-      {% endfor %}
-    </select>
+        {% for editor in editors %}
+          <option value="{{ editor.name }}"
+                  {% if judgment.assigned_to == editor.name %} selected{% endif %}>
+            {{ editor.print_name }}
+          </option>
+        {% endfor %}
+      </select>
+      <input type="submit" class="button-cta" name="assign" value="Assign"/>
```
jacksonj04, line 70: Can we make `Assign` a locale string and unify the language here and on the judgments list?

ds_caselaw_editor_ui/static/js/src/components/judgmentSidebar.js
```diff
@@ -0,0 +1,49 @@
...
+const JudgmentSidebar = {
+  init: function () {
+    const elements = $(".judgment-sidebar");
+    elements.each(function (ix, element) {
+      JudgmentSidebar.initOne(element);
+    });
+  },
+
+  initOne: function (element) {
+    JudgmentSidebar.initAssignForm(element);
```
jacksonj04, line 12: Probably a good answer I haven't had enough coffee to grok; what's the logic of having `initOne` instead of `init` just calling `JudgmentSidebar.initAssignForm` directly?

ds_caselaw_editor_ui/static/js/src/components/judgmentSidebar.js
```diff
@@ -0,0 +1,49 @@
...
+  },
+
+  initAssignForm: function (element) {
+    const form = $(element).find(".judgment-sidebar__judgment-assign-form");
+    const select = form.find("select");
+    const submit = form.find("input[type='submit']");
+    submit.hide();
+    select.on("change", function (event) {
+      console.log("changed");
```
jacksonj04, line 21: Is this leftover debug code? If not, can it be a more explicit message?

ds_caselaw_editor_ui/static/js/src/components/judgmentSidebar.js
```diff
@@ -0,0 +1,49 @@
...
+      $.ajax({
+        type: "POST",
+        url: action,
+        data: form.serialize(),
+        success: function (data) {
+          const successMessage = $(
+            "<div class='context-notification--success judgment-sidebar__context-notification' aria-busy='false' />"
+          );
+          successMessage.text(data["message"]);
+          loading.replaceWith(successMessage);
```
jacksonj04, line 42: Could we make this success message disappear after a few seconds? It keeps loitering around looking like an interactive element.

judgments/tests/test_judgments.py
```diff
@@ -720,3 +717,32 @@ def test_judgment_unpublish_success_view(self, mock_judgment):
...
+        self.client.force_login(User.objects.get_or_create(username="testuser")[0])
+        User.objects.get_or_create(username="testuser2")
+        self.client.post(
+            "/assign", {"judgment_uri": "/test/judgment", "assigned_to": "testuser2"}
+        )
+        api_client.set_property.assert_called_with(
+            "/test/judgment", "assigned-to", "testuser2"
+        )
+
+    @patch("judgments.views.button_handlers.api_client")
```
jacksonj04, line 734: This feels like a really weird case (I'm guessing legacy behaviour from the judgment list) we should try to get rid of if we can, but probably not in this PR.

judgments/views/judgment_hold.py
```diff
@@ -28,6 +29,7 @@ def get_context_data(self, **kwargs):
                 ),
                 "view": "hold_judgment",
                 "judgment": judgment,
+                "editors": editors_dict(),
```
jacksonj04, line 32: We do this in enough places we should definitely just shove this in a context processor so it's always available.

jacksonj04 (changes requested):
A couple of bits aside from the inline comments:

- Can we move this to the sidebar in all cases, including the edit metadata page, so that it's consistent in all contexts?
- The success message feels very large and looks a lot like a button. We might need to revisit the styling on this.

Terry-Price (changes requested):
Hey Tim,

Can we change the term from 'judgment' to 'document' This generic term will be better for  none judgment documents.

ds_caselaw_editor_ui/templates/includes/judgment_sidebar.html
timcowlishaw: done!

ds_caselaw_editor_ui/static/js/src/components/judgmentSidebar.js
timcowlishaw: done!

judgments/tests/test_judgments.py
timcowlishaw: agree, will have a think about how to refactor and do a separate PR

timcowlishaw:
Hey folks @jacksonj04 @Terry-Price , i've responded to changes requested and left some replies to comments, let me know if you're happy for this to go through!

Terry-Price (changes requested):
can we change 'judgment' to 'document' on the notification alert.

timcowlishaw:
> can we change 'judgment' to 'document' on the notification alert.

done, @Terry-Price !

Terry-Price (commented):
i couldnt see the notification message to say document, but assume that’s fine because you mentioned it Tim on here. 

Not for now, but wondered if there is another refactoring card to come out of this for ‘judgment’ class names to be ‘document’ instead. e.g. ‘judgment-cta’ and ‘judgment- sidebar’ and a few others in this space. I may not have recalled the exact class names here, as I can’t see them as I type this.

timcowlishaw:
> i couldnt see the notification message to say document, but assume that’s fine because you mentioned it Tim on here.
> 
> Not for now, but wondered if there is another refactoring card to come out of this for ‘judgment’ class names to be ‘document’ instead. e.g. ‘judgment-cta’ and ‘judgment- sidebar’ and a few others in this space. I may not have recalled the exact class names here, as I can’t see them as I type this.

aha yep Terry, good point - I think we should also trawl the translations for any reference to judgments too as there's doubtless still some user-facing mentions. Best tackling that all at once though than trying to do it here.

jacksonj04 (approved):

jacksonj04:
I'm happy with this, just needs @Terry-Price to green-light it.

PR #842, opened 2023-04-04
use general button styles for judgment toolbar

timcowlishaw:
## Changes in this PR:

Use the existing site-wide button styles for the judgment header toolbar (and adapt them where needed - in particualar we've now added a 'disabled button' style globally.

https://trello.com/c/GccWXdAq/713-eui-ensure-toolbar-buttons-are-underlined-with-correct-double-boarder-on-hover-etc

## Screenshots of UI changes:

imagen


Terry-Price (approved):
looking good to me Tim

PR #852, opened 2023-04-06
Ensure we can unhold judgments

dragon-dxw:
The HTML template for the index assumed `judgment/hold` took a variable to determine whether to hold or unhold, but we have separate hold/unhold pages, so refactored it to use the existing pages.

TODO: item.meta.editor_hold should probably be a true boolean.


jacksonj04 (approved):
I'm happy with this as a patch, but this is definitely smelly. Not easy to fix until result lists use the nice `Judgment` model though, so not worth stressing over.

PR #865, opened 2023-04-12
Invalidate editor and assets caches too

dragon-dxw:
## Changes in this PR:
We discovered that PDFs which had been published and unpublished were not being invalidated. We add blanket invalidation for the editor (`/*`) to bring it in line with the public website and targetted invalidation of the assets (`/ewca/ch/2023/12345/*`).

https://my.slack.com/archives/D04CDM9J2P5/p1681286013046449 [private chat]

# Requires env variable(s) to be updated -- **YES**

`CLOUDFRONT_ASSETS_DISTRIBUTION_ID`
`CLOUDFRONT_EDITOR_DISTRIBUTION_ID` -- names of the distributions can be found at https://eu-west-2.console.aws.amazon.com/cloudfront/v3/home?region=eu-west-2#/distributions (depending on whether logged into staging or prod)


jacksonj04 (approved):
Nice. Any chance of a test to confirm this tries to hit all three things?

dragon-dxw:
https://my.slack.com/archives/C04FQ3GFGUQ/p1681315623387199
Nick: briefly; what's the logic for only the Assets invalidation having a URI?
Dragon: Pages on Public/Editor can be impacted despite not having a shared URI stub, so we are overly aggressive; that is not true of S3. (e.g. the main page may contain details of a published/unpublished judgment)

PR #870, opened 2023-04-12
add time class and timedate

Donna-H:
https://trello.com/c/NQ8zVIAi/765-%F0%9F%94%8D-pui-add-time-tags-where-we-display-datetimes
### Before
before-time-added-eui

### After
after-add-time-class-eui


ds_caselaw_editor_ui/templates/includes/judgment_metadata_panel.html
```diff
@@ -15,7 +15,7 @@
     </li>
     <li>
       <span class="judgment-metadata-panel__key">{% translate "judgment.judgment_date" %}</span>
-      <span class="judgment-metadata-panel__value">{{ judgment.judgment_date_as_string }}</span>
+      <time class="judgment-metadata-panel__value" datetime="YYYY-MM-DD">{{ judgment.judgment_date_as_string }}</time>
```
dragon-dxw, line 18: I think the datetime should be a date in YYYY-MM-DD format, so should be a variable of some description:
https://www.w3schools.com/tags/att_time_datetime.asp

This should probably be `judgment.judgment_date_as_date`

Donna-H: okay @dragon-dxw tomorrow @timcowlishaw shall we look at this also as you mentioned in discussion clarity if we use judgment.judgment_date_as_iso

timcowlishaw (approved):

jacksonj04:
We need to be wary of `judgment_date_as_string`, because that's exactly where we'd change how we render the date if we wanted it in a different format.

This should either have a new class method for spitting out the date as an ISO8601 string, or a test that `<time>` is as expected, or ideally both.

PR #877, opened 2023-04-14
Remove Judgment class from EUI

jacksonj04:
The functionality of this class is now moved to the API library, which means that EUI should be using that instead (so we have fewer places to maintain the functionality).

In an ideal world the API client would be the place which handles initialising a connection object and then passing it around, but to try keep the size of the individual changes down this is now done with a new `get_judgment_by_uri` utils method. Eventually this will refactor out.


ds_caselaw_editor_ui/templates/judgment/edit.html
```diff
@@ -16,7 +16,7 @@ <h4>{% translate "judgments.sidebar.downloads" %}</h4>
         </li>
       {% endif %}
       <li>
-        <a href="{{ judgment.xml_url }}">{% translate "judgment.download_xml" %}</a>
+        <a href="{% url 'full-text-xml' judgment.uri %}">{% translate "judgment.download_xml" %}</a>
```
anthonyhashemi, line 19: Why doe we need `full-text-xml` here but not in the line above?

`<a href="{{ judgment.pdf_url }}">{% translate "judgment.download_pdf" %}</a>`

anthonyhashemi (approved):
Looks good - I agree with the minimal change approach 👍

ds_caselaw_editor_ui/templates/judgment/edit.html
jacksonj04: It's on the (long) list of things I want to make more consistent!

PR #889, opened 2023-04-19
Fix metadata input panel styles

Terry-Price:
## Changes in this PR:
Improved  metadata input text panel with... 
- consistent styling
- replaced Ariel font with open sans
- made the input text bigger to 1rem 
- increased padding to give the input text some air 
- made the borders all 2px. 
- reduced the size of NCN field to reflect likely content length.
- increased Court name field to reflect likely content length.
- fixed the Name text field from being draggable.

https://trello.com/c/lQEnfsXZ/812-eui-refine-metadata-panel-styling
### Before
Screenshot 2023-04-19 at 09 38 56

### After
Screenshot 2023-04-19 at 09 38 40

- [ ] 

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_metadata.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_metadata.scss
@@ -24,11 +24,11 @@
   }
 
   &__ncn {
-    width: 40%;
+    width: 27%;
   }
 
   &__court {
-    width: 40%;
+    width: 52%;
     left: 3rem;
   }
   &__date {
@@ -107,12 +107,17 @@
     box-sizing: border-box;
     height: calc(100% - 1.45rem);
     font-size: 1rem;
+    resize: none;
   }
 
   input[type=text] {
     box-sizing: border-box;
     width: 100%;
     min-height: 2rem;
+    font-size: 1rem;
+    $font__open-sans: 'Open Sans', sans-serif;
+    padding: 0.5rem;
+    border: 2px solid $color__dark-grey;
   }
 
   input[type=submit] {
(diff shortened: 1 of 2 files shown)


ds_caselaw_editor_ui/templates/pages/style_guide.html
```diff
@@ -4,17 +4,18 @@
   <div class="style-guide">
     <h2>{% translate "style_guide.title" %}</h2>
     <p>{% translate "style_guide.description" %}</p>
-    <hr />
```
anthonyhashemi, line 7: @Terry-Price do we need to update the base branch to be `style-guide/horizonal-rules` so that we don't see the extra commit from that one until thats merged in?

Terry-Price: @anthonyhashemi Not sure, this one and the hr rules are not dependants. happy happy with whatever is usual.

anthonyhashemi: Ah okay - well if the other gets merged in first then you dont need to do anything so I wouldnt worry.

ds_caselaw_editor_ui/sass/includes/_metadata.scss
```diff
@@ -24,11 +24,11 @@
   }
```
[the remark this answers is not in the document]
jacksonj04: I think we run Ghost Inspector on production, but I don't know if we do on EUI. @dragon-dxw may know more.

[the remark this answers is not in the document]
timcowlishaw: aha we don't currently do any of that, but we probably should. Maybe one to think about more generally and add to the front end tech debt column in trello!

timcowlishaw (approved):

PR #897, opened 2023-04-21
Tidying in advance of prettifying judgment versions

jacksonj04:
We want to add the ability to view parser logs.

The sensible place for this is by the versions list.

The versions list does not look pretty.

We want to make it look pretty, but there was stuff in the way which distracted me.

This PR fixes that stuff.


ds_caselaw_editor_ui/sass/main.scss
```diff
@@ -20,10 +19,7 @@
 @import "includes/judgment_text_source";
 @import "includes/judgment_sidebar";
 @import "includes/judgment_metadata_panel";
-@import "includes/judgment_text_service_introduction";
```
anthonyhashemi, line 23: my css is rust, how do we know these werent being used?

locale/en_GB/LC_MESSAGES/django.po
```diff
@@ -62,143 +54,165 @@ msgstr "Something wrong with this page?"
...
 msgid "judgment.judgment_name"
 msgstr "Name"
 
-#: ds_caselaw_editor_ui/templates/includes/judgment_sidebar.html
+#: ds_caselaw_editor_ui/templates/includes/judgment/metadata_panel_form.html
+#: ds_caselaw_editor_ui/templates/includes/judgment/metadata_panel_static.html
+#: ds_caselaw_editor_ui/templates/includes/judgment/sidebar.html
+#: ds_caselaw_editor_ui/templates/includes/judgments_list_item.html
+#: judgments/views/judgment_edit.py
+msgid "judgments.consignmentref"
```
anthonyhashemi, line 92: why have we moved these blocks around? Are they now ordered in a more organised way?

jacksonj04: It's because this file is autogenerated; they're sorted by the first file where the string is encountered in a depth-first search of an alphabetically sorted tree. We renamed some files so the blocks get moved, and the same thing will happen any time the translations are updated (so we may as well do it here and reduce future diffs).

It's not my favourite feature of Django's translations.

anthonyhashemi: makes complete sense - I havent done much with django translations, so thanks for explaining

ds_caselaw_editor_ui/sass/main.scss
jacksonj04: The "service introduction" stuff has leaked over from PUI when the CSS was copied; it's the intro banner.

The "source" stuff also leaked over from PUI when we had a "this judgment has come from X" at the bottom of each judgment, but this entire feature has now been parked.

PR #927, opened 2023-05-03
Initial version of update NCN and version merging workflow

dragon-dxw:
This provides a new workflow for editors to update the NCN of a judgment (aka the "move" workflow). Updating an NCN will also update the URI of a judgment, and making editors use this workflow provides an opportunity to more fully explain the consequences of doing so.

This workflow will also (in a somewhat hidden way) let you append one document to another as an updated version (aka the "merge" workflow). This is always available if a document is given the same NCN as another, but this functionality isn't advertised.

Upcoming changes to the ingester will allow documents in this 'colliding' state to be surfaced and editors offered the "merge" workflow.

### DIFF
--- a/judgments/utils/__init__.py
+++ b/judgments/utils/__init__.py
@@ -24,6 +24,10 @@ class MoveJudgmentError(Exception):
     pass
 
 
+class OverwriteJudgmentError(Exception):
+    pass
+
+
 class NeutralCitationToUriError(Exception):
     pass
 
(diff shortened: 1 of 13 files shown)


dragon-dxw:
[Invent a process for merging two discrete documents in ML](https://trello.com/c/H2KIIobf/782-invent-a-process-for-merging-two-discrete-documents-in-ml)

dragon-dxw:
To do: needs to move docx files from old/uri/old_uri.docx to new/uri/new_uri.docx instead of just new/uri/old_uri.docx

anthonyhashemi:
Should I hold off reviewing this until its been tested? @dragon-dxw @jacksonj04

judgments/utils/__init__.py
```diff
@@ -63,6 +67,50 @@ def get_judgment_root(judgment_xml) -> str:
...
+        old_judgment_bytes = old_judgment.content_as_xml()
+        old_judgment_xml = ET.XML(bytes(old_judgment_bytes, encoding="utf-8"))
+        api_client.save_judgment_xml(
+            new_uri,
+            old_judgment_xml,
+            annotation=f"overwritten from {old_uri}",
+        )
+        set_metadata(old_uri, new_uri)
+        # TODO: consider deleting existing public assets at that location
+        copy_assets(old_uri, new_uri)
```
dragon-dxw, line 97: I was concerned that pdf/docx files would not be renamed, but utils.aws.copy_assets already does this -- this is the same handler as the existing approach.

anthonyhashemi:
Should we close this PR for now to reduce the number of open PRs given that this work is still not past design stage?

PR #931, opened 2023-05-04
FCL 29: refactor tests to support judgment exception

dragon-dxw:
A number of editor tests fail because we've changed how we check if a judgment is present, so the mocks were wrong.

**REQUIRES** v7 of the custom-api-client

Ready to review, but **not** ready to merge!

### DIFF
--- a/judgments/tests/test_judgments.py
+++ b/judgments/tests/test_judgments.py
@@ -3,7 +3,6 @@ from unittest.mock import patch
 from urllib.parse import urlencode
 
 from django.contrib.auth.models import User
-from django.http import Http404
 from django.test import TestCase
 from django.urls import reverse
 from django.utils.translation import gettext
@@ -267,7 +266,6 @@ class TestJudgmentView(TestCase):
         )
         self.assertEqual(response.status_code, 404)
 
-
     def test_judgment_xml_view_redirect(self):
         self.client.force_login(User.objects.get_or_create(username="testuser")[0])
         response = self.client.get("/xml?judgment_uri=ewca/civ/2004/63X")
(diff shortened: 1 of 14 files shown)


judgments/tests/test_judgments.py
```diff
@@ -487,7 +487,7 @@ def test_judgment_hold_view_invalid_uri(self, mock_api_client):
                 "judgment_uri": "invalid",
             },
         )
-        assert response.status_code == 400
+        assert response.status_code == 404
```
jacksonj04, line 481: Amazed that these weren't 404s to begin with, nice tweak to make the tests reflect what should really be happening.

judgments/tests/test_judgments.py
```diff
@@ -229,8 +229,8 @@ def test_judgment_html_view_with_failure(self, mock_judgment):
         assert response.status_code == 200
 
     @patch(
-        "caselawclient.models.judgments.MarklogicApiClient.get_judgment_citation",
-        side_effect=MarklogicResourceNotFoundError(),
+        "caselawclient.models.judgments.Judgment.judgment_exists",
```
jacksonj04, line 231: These definitely should be changed (because I think the original mock wasn't in the right place), but I think they should be changed to mock the side effect we expect from `get_judgment` so that we're actually testing the view behaves as expected (ie we're not testing the implementation of `get_judgment` and the `Judgment` model as well.

dragon-dxw:
This feels like it got out of hand, so might be worth pairing on Tuesday to get this over the line.

jacksonj04 (approved):

PR #947, opened 2023-05-15
Fix encoding issues in autogenerated email texts

jacksonj04:
The generated links for emails are incorrectly encoding special HTML characters like `&`, `<`, `>`, `'` and `"`. Fix the layers at which encoding happens so these are properly encoded for links.


judgments/utils/external.py
```diff
@@ -0,0 +1,44 @@
+from urllib.parse import quote, urlencode
```
anthonyhashemi, line 1: I feel like `email_links` could be a more informative file name. What do you think?

judgments/utils/external.py
```diff
@@ -0,0 +1,44 @@
...
+
+from django.template import loader
+
+
+def build_email_link_with_content(address, subject, body=None):
+    """Given a destination address, subject, and (optionally) body for an email build a mailto link for it."""
```
anthonyhashemi, line 7:
```suggestion
    """Given a destination address, subject, and (optionally) body for an email, build a mailto link for it."""
```

judgments/utils/external.py
```diff
@@ -0,0 +1,44 @@
...
+    body_string = loader.render_to_string(
+        "emails/confirmation_to_submitter.txt", email_context
+    )
+
+    return build_email_link_with_content(
+        judgment.source_email, subject_string, body_string
+    )
+
+
+def build_raise_issue_email_link(judgment):
```
anthonyhashemi, line 39: can we have a test for this?

judgments/utils/external.py
```diff
@@ -0,0 +1,44 @@
...
+
+    if body:
+        params["body"] = body
+
+    return "mailto:{address}?{params}".format(
+        address=address, params=urlencode(params, quote_via=quote)
+    )
+
+
+def build_confirmation_email_link(request, judgment):
```
anthonyhashemi, line 18: can we have a test for this?

judgments/tests/utils/test_external.py
```diff
@@ -0,0 +1,31 @@
...
+            )
+            == "mailto:test@example.com?subject=Find%20Case%20Law%20%E2%80%93%20FAO%3A%20Email%20Subject&body=Dear%20Test%2C%0A%0AThis%20is%20an%20email."  # noqa: E501
+        )
+
+    def test_build_email_link_with_content_with_special_chars(self):
+        assert (
+            build_email_link_with_content(
+                "test@example.com",
+                "Subject",
+                """Ampersand: & Quote: ' Double Quote: " LT: < GT: >""",
```
anthonyhashemi, line 28: nice 👍

judgments/views/judgment_edit.py
```diff
@@ -18,51 +18,14 @@
...
-    def build_raise_issue_email_link(self, context):
-        subject_string = "Issue(s) found with {reference}".format(
-            reference=context["judgment"].consignment_reference
-        )
-
-        return build_email_link_with_content(
-            context["judgment"].source_email, subject_string
-        )
-
     def build_jira_create_link(self, request, context):
```
anthonyhashemi, line 29: Can we move this to the utils file too to be consistent?

In that case my email_links suggestion won't be quite accurate anymore,

maybe `link_builders`?

ds_caselaw_editor_ui/templates/emails/confirmation_to_submitter.txt
```diff
@@ -1,6 +1,6 @@
 Dear XXXXXX,
 
-We are pleased to inform you that your uploaded judgment "{{ judgment_name }}", reference {{ reference }}, has been published on Find Case Law. You can see the published judgment at {{ public_judgment_url }}.
+We are pleased to inform you that your uploaded judgment "{{ judgment_name|safe }}", reference {{ reference }}, has been published on Find Case Law. You can see the published judgment at {{ public_judgment_url|safe }}.
```
anthonyhashemi, line 3: Marking these inputs as safe is the only fix here, right?

dragon-dxw: We could wrap the whole thing in `{% autoescape off %}` https://docs.djangoproject.com/en/4.2/ref/templates/language/#how-to-turn-it-off; but these are broadly equivalent.

jacksonj04: Yeah, I pondered both but I'd rather be explicit about things rather than just wrap the entire template.

The only other alternative is a custom template renderer, which seems a lot of extra work.

anthonyhashemi (approved):
Nice one 😄

PR #956, opened 2023-05-23
Mark email fields as safe and remove repetition in email header

dragon-dxw:
The text of the email must be interpretted without escaping, as there is no process that
will unescape it, unlike in a browser context.

Addresses https://trello.com/c/baGbbWHc/922-eui-changing-to-the-wording-of-email-from-eui-needed

We also remove a doubling up of the subject name -- Find Case Law is automatically prepended

### DIFF
--- a/ds_caselaw_editor_ui/templates/emails/confirmation_to_submitter.txt
+++ b/ds_caselaw_editor_ui/templates/emails/confirmation_to_submitter.txt
@@ -1,4 +1,4 @@
-Dear {{ submitter|safe}},
+Dear {{ submitter|safe }},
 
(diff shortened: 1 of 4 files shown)


dragon-dxw:
[EUI: Changing to the wording of email from EUI needed](https://trello.com/c/baGbbWHc/922-eui-changing-to-the-wording-of-email-from-eui-needed)

ds_caselaw_editor_ui/templates/emails/confirmation_to_submitter.txt
```diff
@@ -1,12 +1,12 @@
-Dear {{ submitter }},
+Dear {{ submitter|safe}},
```
anthonyhashemi, line 1: can we add tests asserting the correct strings in the rendered template?

dragon-dxw: Added.

anthonyhashemi (approved):
Nice one 👍

PR #995, opened 2023-06-15
Tidy up (some) container behaviour

jacksonj04:
The edit judgment page has a lot of conflicting concepts about how wide things should be; simplify this by removing some nested concepts of container, and standardise container widths to match PUI.


timcowlishaw (approved):
A vast improvement, thank you!

PR #1007, opened 2023-06-23
Fix for missing metadata in list of unpublished judgments

jacksonj04:
The new unified `SearchResult` class from the API Client uses `SearchResult.metadata` instead of `SearchResult.meta`, and this change was missed here.

This fixes both the missing metadata, and the broken assignment operation.


anthonyhashemi (approved):
Looks good!

PR #1017, opened 2023-06-29
Updating judgment uri sets collection based off new uri

anthonyhashemi:
## Changes in this PR:

- Update marklogic api client to 11.0.0 https://github.com/nationalarchives/ds-caselaw-custom-api-client/releases/tag/v11.0.0 to make use of the updated and renamed `copy_document` method so that it sets the collection of the new document based off the uri, so that we dont get floating documents which were at a /failure/ uri since the parser failed to pick up an ncn which existed in the document and then got updated to a real ncn uri by an editor.

Without this, these docs that were published were not showing up in search due to the fact that we were creating new docs without a collection by when copying a document, and we filter so that we only show docs in judgment collection.

Note: the reason failure uris show in judgment collection is that we default to anything without press-summary in the uri to be in the judgment collection for now.

https://trello.com/c/0Q5fI6gA/1072-ingesting-uri-parse-failures-not-being-added-to-judgment-or-press-summary-collection

### DIFF
--- a/judgments/tests/utils/test_utils.py
+++ b/judgments/tests/utils/test_utils.py
@@ -15,7 +15,7 @@ from judgments.utils import (
     extract_version,
     get_judgment_root,
     render_versions,
-    update_judgment_uri,
+    update_document_uri,
 )
 from judgments.utils.aws import build_new_key, invalidate_caches
 from judgments.utils.paginator import paginator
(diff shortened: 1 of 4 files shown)


dragon-dxw:
> Note: the reason failure uris show in judgment collection is that we default to anything without press-summary in the uri to be in the judgment collection for now.

This gave me a fright but it's fine -- they won't be *published* judgments.

dragon-dxw:
ds-caselaw-editor-ui/judgments/tests/utils/test_utils.py also refers to copy_judgment (hence the failing tests)

dragon-dxw (changes requested):
ds-caselaw-editor-ui/judgments/tests/utils/test_utils.py also refers to copy_judgment (hence the failing tests)

anthonyhashemi:
> > Note: the reason failure uris show in judgment collection is that we default to anything without press-summary in the uri to be in the judgment collection for now.
> 
> This gave me a fright but it's fine -- they won't be _published_ judgments.

Yep exactly! I have a ticket to look into passing the info of judgment vs press summary outside ncn from the parser so that we can distinguish them for when we want to separate out the editor ui, but for now its not particularly an issue

anthonyhashemi:
> test_utils

Cheers

anthonyhashemi:
> ds-caselaw-editor-ui/judgments/tests/utils/test_utils.py also refers to copy_judgment (hence the failing tests)

Have updated along with some more name updates, but havent touched `judgments/views/judgment_edit.py` as I know @jacksonj04 's  work is touching all of that flow.

dragon-dxw (approved):

PR #1019, opened 2023-06-30
Publication validation messages now only show in publish flow

jacksonj04:
## Changes in this PR:

The blue box containing publication validation messages (ie why a document cannot be published) previously appeared in all contexts of viewing a publication.

It is now visible _only_ on the publication page, but the 'Publish' button is now always enabled so editors can always begin this flow and see any validation errors.

### Before

localhost_3000_ewhc_kb_2023_1593

### After

localhost_3000_ewhc_kb_2023_1593 (1)

localhost_3000_ewhc_kb_2023_1593_publish


anthonyhashemi (approved):
😎

dragon-dxw (approved):

PR #1022, opened 2023-06-30
Remove ability to update NCN using the edit form

dragon-dxw:
Removing this from the edit work since it means we can't deploy safely; we shouldn't deploy this until we're happy.


jacksonj04 (changes requested):
We can't merge this until we have an alternative in place; we're currently validating this with the editors via research.

anthonyhashemi:
Should we close this PR for now to reduce the number of open PRs given that this work is still not past design stage?

PR #1030, opened 2023-07-06
Fix/header padding

jacksonj04:
## Changes in this PR:

Fix the black padding on the Labs button which has been bugging me since we slimmed down the header.

### Before

localhost_3000_

### After

localhost_3000_ (1)


Terry-Price (approved):
nice Nick

PR #1041, opened 2023-07-07
remove "view on public site" button

Donna-H:
## Changes in this PR:
Remove "view on public site" button
https://trello.com/c/36oo8voH/1127-eui-remove-view-on-public-site-button-from-the-tool-bar
### Before
before-public

### After
after-public

### DIFF
--- a/ds_caselaw_editor_ui/templates/includes/judgment/toolbar.html
+++ b/ds_caselaw_editor_ui/templates/includes/judgment/toolbar.html
@@ -58,17 +58,6 @@
         <span class="judgment-toolbar__button button-secondary"
               aria-disabled="true">{% translate "judgment.toolbar.download_docx" %}</span>
       {% endif %}
-      {% if judgment.is_published %}
-        <a class="judgment-toolbar__button button-secondary"
-           href="{{ judgment.public_uri }}"
-           target="_blank"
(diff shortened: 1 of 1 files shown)


Donna-H:
[EUI: remove "view on public site" button from the tool bar](https://trello.com/c/36oo8voH/1127-eui-remove-view-on-public-site-button-from-the-tool-bar)

timcowlishaw (approved):
Ace, thank you donna!

jacksonj04:
I'm unconvinced about this; once the 'Edit' page is no more this will mean there is no way to get from the EUI to the corresponding public view of the document without dropping into a specific workflow (which feels bad to me).

What is the problem with this button continuing to exist?

timcowlishaw:
I think that's a good question, Nick - I had the same doubts - as I
understand it, the only time the editors use it currently in the "typical
workflow" is at the end of the publication flow, where a link still exists,
but im not sure thats a particularly strong argument for removal though, it
seems a weird omission not to have a link for published judgements to the
public version somewhere prominent (although there might be a better place
for it, true). Maybe one to take up with rose on Tuesday?

El ds., 8 jul. 2023, 10.48, Nick Jackson ***@***.***> va
escriure:

PR #1042, opened 2023-07-07
remove email links box in edit page

Donna-H:
## Changes in this PR:
Remove email links box in edit page
https://trello.com/c/bBrGOAJB/1136-eui-remove-the-yellow-email-links-box-from-the-edit-page
### Before
before-email-block

### After
after-email-block


Donna-H:
[EUI: Remove the yellow 'email links' box from the edit page](https://trello.com/c/bBrGOAJB/1136-eui-remove-the-yellow-email-links-box-from-the-edit-page)

timcowlishaw (approved):

jacksonj04 (changes requested):
This email link block is also used in the publish judgment flow, so we can't remove the template entirely.

Donna-H:
> This email link block is also used in the publish judgment flow, so we can't remove the template entirely.

@jacksonj04 @timcowlishaw  the email box removed was only removed on edit page a different version of the email links box still remains on both hold and publish pages see below:
published-box
onholdbox

jacksonj04 (approved):
Ooh, my bad! I misread `email_confirm_publication.html` as `create_emails_block.html`.

Looks good to merge to me, my only other comment would be that the email block we delete makes use of translation strings, so we could do with running the `makemessages` step to remove the deprecated strings from the translations file.

Donna-H:
> We should probably remove the code which generates this link as well!

@jacksonj04 just checking is it this one you added to be deleted "def build_raise_issue_email_link" in judgments/utils/link_generators.py ?

jacksonj04:
@Donna-H almost! We actually use that bit of code in two places - it's used in the [edit page](https://github.com/nationalarchives/ds-caselaw-editor-ui/blob/remove-email-links-box/judgments/views/judgment_edit.py#L20), but also the [hold page](https://github.com/nationalarchives/ds-caselaw-editor-ui/blob/remove-email-links-box/judgments/views/judgment_hold.py#L11).

What we actually just need to do is remove [the bits where we drop `email_raise_issue_link` and `email_confirmation_link` into the context of the page](https://github.com/nationalarchives/ds-caselaw-editor-ui/blob/remove-email-links-box/judgments/views/judgment_edit.py#L39-L44), and their associated imports (see [this commit](https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/1042/commits/12151e2361a637e86a8ad6cb974a6b211fe72307)).

PR #1045, opened 2023-07-07
remove checkboxes on edit page

Donna-H:
## Changes in this PR:
Remove checkboxes on edit page

https://trello.com/c/y9HWdgtn/1137-eui-remove-the-three-problems-checkboxes-from-the-edit-page
### Before
before-chekboxes

### After
after-checkboxes-removed


Donna-H:
[EUI: Remove the three 'problems' checkboxes from the edit page](https://trello.com/c/y9HWdgtn/1137-eui-remove-the-three-problems-checkboxes-from-the-edit-page)

timcowlishaw (approved):

jacksonj04 (changes requested):
If we're definitely losing these features, can we also remove the relevant code in the view (and probably the API and database), so we don't have dead code?

Shout up if you need a hand!

Donna-H:
> If we're definitely losing these features, can we also remove the relevant code in the view (and probably the API and database), so we don't have dead code?
> 
> Shout up if you need a hand!

@jacksonj04 @timcowlishaw the box was only removed from edit page as per the Trello ticket a different version of the box 
 still remains in publish page and also hold page screenshots below:
after-email-block
onholdbox
published-box

Donna-H:
@jacksonj04 @timcowlishaw  I will write up a new Trello Backend card for the API and Database and link it on Frontend card

dragon-dxw (approved):

jacksonj04 (approved):

PR #1067, opened 2023-07-12
Added status and type label to EUI

Donna-H:
## Changes in this PR:
Added status and type label as per design here https://xsodca.axshare.com/?id=ae2k60&p=metadata-panel-with-status-and-type&c=1 some backend required BE placeholder after Type label see images below card has been created for BE.

https://trello.com/c/ucQ9ia8u/1134-eui-add-judgment-status-and-doc-type-label-into-metadata-panel-fe-dev
### Before
before

### After
Judgment
judgment-view

Press Summary
press-summary-view

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_metadata.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_metadata.scss
@@ -19,39 +19,77 @@
     margin-bottom: calc($spacer__unit / 2);
     &:last-child {
       margin-bottom: 0;
+      padding: calc($spacer__unit / 2) 0 calc($spacer__unit / 2) 0;
     }
   }
 
   &__ncn {
-    width: 27%;
+    width: 18%;
   }
 
   &__court {
     width: 52%;
-    left: 3rem;
   }
+
   &__date {
-    width: 20%;
+    width: 15%;
   }
+
   &__tdr {
+    display: flex;
+    gap: 1rem;
     p {
       margin: 0;
+      margin-left: 0.5rem;
       line-height: 1rem;
     }
     margin-bottom: calc($spacer__unit / 2);
   }
+
+  &__type {
+    display: flex;
+    gap: 1rem;
+    margin-left: 0.5rem;
+    p {
+      margin: 0;
+      margin-left: 1.5rem;
+      line-height: 1rem;
+    }
+    margin-bottom: calc($spacer__unit / 2);
+  }
+
   &__panel {
     width: 80%;
+    padding: calc(#{$spacer__unit} / 2) 0;
   }
 
   &__bottom-right-column {
     width: 20%;
     padding: calc($spacer__unit / 2) 0;
+    margin-left: calc($spacer__unit / 0.5);
+  }
+
+  &__top-right-column {
+    width: 20%;
   }
 
   &__actions {
     width: auto;
   }
+
+  &__status {
+    display: flex;
+    gap: 1rem;
+    p {
+      margin: 0;
+      margin-left: 1rem;
+      line-height: 1rem;
+      border: 3px solid white;
+    }
+    margin-bottom: calc($spacer__unit / 2);
+    width: auto;
+  }
+
   &__main-labels {
     display: block;
     margin-bottom: calc($spacer__unit / 4);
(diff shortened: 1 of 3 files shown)


Donna-H:
[EUI - Add judgment status and doc type label into metadata panel [FE DEV]](https://trello.com/c/ucQ9ia8u/1134-eui-add-judgment-status-and-doc-type-label-into-metadata-panel-fe-dev)

ds_caselaw_editor_ui/templates/includes/judgment/metadata_panel_form.html
```diff
@@ -51,11 +67,11 @@
...
         </div>
         <div class="metadata-component__bottom-right-column">
-          <div class="metadata-component__tdr">
+          <div class="metadata-component__type">
             <aside for="metadata_name" class="metadata-component__main-labels">
-              {% translate "judgments.consignmentref" %}
+              {% translate "document.type" %}
             </aside>
-            <p>{{ judgment.consignment_reference }}</p>
+            <p>Judgment or Press Summary BE</p>
```
anthonyhashemi, line 74: for the backend portion we plan on adding a `document_type` property to the Judgment / Press summary / Document object.

So if you want to prepare for that you can update the template to read in from a context variable, say `{{ document_type }}` here, which will be defined by `judgment.document_type` in the view. 

Then once we have updated the apiclient and released a new version, you will just have to update the EUI to use the new version in this PR.

In the meantime, if you dont want to be completely blocked and get your PR in for now, you could always utilise the same logic we did in the PUI:


```
    press_summary_suffix = "/press-summary/1"
    if document_uri.endswith(press_summary_suffix):
        context["document_type"] = "press_summary"
    else:
        context["document_type"] = "judgment"
```

from `judgments/views/detail.py` in the PUI.

Donna-H: @anthonyhashemi updated`{{ judgment.document_type }}` have other FE in the 'To Do' working on these now. So this PR can remain in blocked once BE have updated the apiclient and released a new version,  new card for updating the EUI to use the new version in this PR.

ds_caselaw_editor_ui/sass/includes/_metadata.scss
```diff
@@ -19,39 +19,77 @@
     margin-bottom: calc($spacer__unit / 2);
     &:last-child {
       margin-bottom: 0;
+      padding: calc($spacer__unit / 2) 0 calc($spacer__unit / 2) 0;
     }
   }
 
   &__ncn {
```
timcowlishaw, line 26: One for the refactoring column rather than a thing to do now, but it might be nice to refactor here to use Flexbox or grid, so we don't have to approximate and hardcode these percentages

timcowlishaw (changes requested):
Looks great to me Donna! one very minor visual nitpick - would it be possible to take a quick look at what's going on with the spacing between the TDR Ref, status and document type? it looks like there's about half as much space between the first two as the second, and the "status" label doesn't quite line up vertically with the actual status.

Thank you!

252953813-9a717b34-48e7-4008-aede-2adaed665e62

timcowlishaw:
> Looks great to me Donna! one very minor visual nitpick - would it be possible to take a quick look at what's going on with the spacing between the TDR Ref, status and document type? it looks like there's about half as much space between the first two as the second, and the "status" label doesn't quite line up vertically with the actual status.
> 
> Thank you!
> 
> 252953813-9a717b34-48e7-4008-aede-2adaed665e62


Aha Donna, don't worry about this - I'm doing a bit of refactoring and reckon it'll be easier to look at at the same time! If it's ok with you i'll add a couple of tweaks to this branch myself before we merge.

timcowlishaw:
@Donna-H I've pushed a couple of tweaks, I hope you don't mind - I'll go through them with you when you're back before merging!

timcowlishaw (approved):

ds_caselaw_editor_ui/templates/includes/judgment/metadata_panel_form.html
anthonyhashemi: The actual judgment property might be slightly different, but once @jacksonj04 has done that work we can update this appropriately

PR #1071, opened 2023-07-17
Improve pre-commit hooks configuration

jacksonj04:


anthonyhashemi (approved):

dragon-dxw:
[PUI: Family Court filter - displaying 'b' when it should not be](https://trello.com/c/FRTNXptE/1629-family-court-filter-displaying-b-when-it-should-not-be)

PR #1082, opened 2023-07-24
Add a hover color to the tabset, and fix a bug when used twice on the same page

timcowlishaw:
## Changes in this PR:

Two minor improvements, we now have a hover colour on unselected tabs (and the cursor displays correctly on selected tabs), and the colours for hover/inactive/selected are consistent with the CTAs (before they were reversed)

I also i fixed a slightly puzzling thing where if the tabset is used twice in the same page, the first instance appears to break (showing all tabs as unselected). This has meant a slight change to the interface - if it's used twice on the same page, each tabset must be given a unique `name` property, which is noted in the styleguide.

https://trello.com/c/0pOIKYIQ/1189-bug-tabs-component-design-fix-hover-colour
### Before
Screenshot 2023-07-24 at 14 22 27

#### No Hover
Screenshot 2023-07-24 at 14 58 30
#### Hover

Screenshot 2023-07-24 at 14 58 35

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_tabs.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_tabs.scss
@@ -7,9 +7,9 @@
     margin: 0 0.25rem;
 
     label {
-      background: $color__dark-blue;
+      background: $color__aqua-blue;
       padding-top: 0.75rem;
-      border: 2px solid $color__dark-blue;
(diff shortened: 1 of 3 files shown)


dragon-dxw (approved):

Terry-Price:
Looking fantasitic @timcowlishaw - thank you so much.

PR #1101, opened 2023-08-07
changed colour of border bottom in static metadata

Terry-Price:
Border bottom should be 'Almost black rather than blue.
Made this change on Sass.
https://trello.com/c/FNLPsMNG/1232-eui-fix-colour-border-bottom-for-static-metadata-panel
### Before
Screenshot 2023-08-07 at 14 47 44

### After
Screenshot 2023-08-07 at 14 46 50


Terry-Price:
[EUI: fix colour border bottom for static metadata panel](https://trello.com/c/FNLPsMNG/1232-eui-fix-colour-border-bottom-for-static-metadata-panel)

jacksonj04 (commented):
I'm not sure why sass is now included as a dependency in the `package.json`, was that intentional?

PR #1102, opened 2023-08-07
Make NCNs and correct document type show in press summary documents' metadata panels

anthonyhashemi:
## Changes in this PR:

- Make NCNs show in press summary documents' metadata panels 
- Make "press summary" show as document type for press summaries

- https://trello.com/c/inj3m6KV/1230-eui-apiclient-bug-no-ncn-showing-for-press-summaries
- https://trello.com/c/xM8Yxtx7/1228-eui-apclient-bug-incorrect-document-type-showing-for-press-summaries

### Before
Screenshot 2023-08-07 at 15 36 34

### After
Screenshot 2023-08-07 at 15 36 02

### DIFF
--- a/ds_caselaw_editor_ui/templates/includes/judgment/metadata_panel_form.html
+++ b/ds_caselaw_editor_ui/templates/includes/judgment/metadata_panel_form.html
@@ -14,7 +14,7 @@
             </label>
             <input type="text"
                    class="metadata-component__neutral_citation-input"
-                   value="{{ judgment.neutral_citation }}"
+                   value="{{ judgment.best_human_identifier }}"
                    name="neutral_citation"
                    id="neutral_citation" />
           </div>
--- a/judgments/tests/test_judgments.py
+++ b/judgments/tests/test_judgments.py
@@ -14,7 +14,7 @@ def assert_match(regex, string):
 
 
 class TestJudgmentView(TestCase):
-    @patch("judgments.views.full_text.get_judgment_by_uri_or_404")
+    @patch("judgments.views.full_text.get_document_by_uri_or_404")
     @patch("judgments.utils.api_client.document_exists")
     @patch("judgments.utils.api_client.get_document_type_from_uri")
     def test_judgment_html_view(self, document_type, document_exists, mock_judgment):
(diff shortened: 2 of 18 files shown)


dragon-dxw (approved):

ds_caselaw_editor_ui/templates/includes/judgment/metadata_panel_form.html
```diff
@@ -14,7 +14,7 @@
             </label>
             <input type="text"
                    class="metadata-component__neutral_citation-input"
-                   value="{{ judgment.neutral_citation }}"
```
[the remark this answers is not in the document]
jacksonj04: Yeah, the fact that all this is using `judgment` instead of `document` is something to solve in the firebreak once PRs are out.

judgments/utils/view_helpers.py
```diff
@@ -46,10 +43,3 @@ def get_search_results(parameters: dict[str, Any]) -> dict[str, Any]:
         "order": parameters["order"],
         "paginator": paginator(parameters["page"], search_response.total),
     }
-
-
-def get_judgment_by_uri_or_404(uri: str) -> Judgment:
-    try:
-        return get_judgment_by_uri(uri)
```
jacksonj04, line 53: Wouldn't it be simpler to swap to using `api_client.get_document_by_uri` here and keep the rest of the structure intact? I'm cool with the extra layer to raise a clean 404 if not.

(Long-term we should absolutely have a superclass for "this view is about a document" so we can ditch a load of boilerplate)

judgments/tests/test_judgments.py
```diff
@@ -14,7 +14,7 @@ def assert_match(regex, string):
 
 
 class TestJudgmentEdit(TestCase):
-    @patch("judgments.views.judgment_edit.get_judgment_by_uri_or_404")
```
jacksonj04, line 17: Doesn't this now cause a `DocumentNotFoundError` to be raised if it doesn't exist rather than a `Http404`?

anthonyhashemi: ah yeah, true, will do what you suggested above

PR #1105, opened 2023-08-08
Fix for document context not always being available.

jacksonj04:
The way we load up document contexts is messy and inconsistent. This PR fixes that:

- Introduce a new `DocumentView` class which is based on the Django `TemplateView` classy view. This means we can easily take advantage of Django's handling of template loading and rendering (which reduces our boilerplate), and only deal with loading document context in one place (which cuts down duplication and reduces scope for error).
- Do a whole bunch of renaming to reflect that we're actually handling `Document`s and not exclusively `Judgment`s.


judgments/utils/view_helpers.py
```diff
@@ -53,3 +61,31 @@ def get_judgment_by_uri_or_404(uri: str) -> Judgment:
...
+
+        context["document"] = document
+
+        # TODO: Remove this once we fully deprecate 'judgment' contexts
+        context["judgment"] = document
+
+        context["page_title"] = document.name
+        context["courts"] = caselawutils.courts.get_all()
+
+        context.update({"editors": editors_dict()})
```
anthonyhashemi, line 83: can we simplify this to

```suggestion
        context["editors"] =  editors_dict()
```

judgments/views/document_history.py
```diff
@@ -0,0 +1,5 @@
+from ..utils.view_helpers import DocumentView
+
+
+class DocumentHistoryView(DocumentView):
+    template_name = "judgment/history.html"
```
dragon-dxw, line 5: I'm not seeing this file, is it committed if it's new?
Probably me being an idiot.

dragon-dxw (approved):

judgments/views/document_history.py
jacksonj04: It's at https://github.com/nationalarchives/ds-caselaw-editor-ui/blob/main/ds_caselaw_editor_ui/templates/judgment/history.html

PR #1106, opened 2023-08-08
Expose press summaries in search results

anthonyhashemi:
Currently we filter out docs so that only docs in the judgment collection are shown in the EUI search results, to be in line with the PUI where we have decided not to show press summaries in search results.

This doesn’t really make sense in EUI, as we want to see all unpublished judgments in the EUI (even though editors do normally just work from Jira and not the unpublished list/ search results)

## Changes in this PR:
- Expose press summaries in search results

https://trello.com/c/GlNGCcIC/1240-eui-expose-press-summaries-in-eui-search-results

### Before
Screenshot 2023-08-08 at 17 09 49

### After
Screenshot 2023-08-08 at 17 09 24

### DIFF
--- a/judgments/tests/test_judgment_list.py
+++ b/judgments/tests/test_judgment_list.py
@@ -13,7 +13,7 @@ def assert_match(regex, string):
 
 
 class TestJudgmentView(TestCase):
-    @patch("judgments.utils.view_helpers.search_judgments_and_parse_response")
+    @patch("judgments.utils.view_helpers.search_and_parse_response")
     def test_judgment_list_smoketest(self, mock_search_results):
         mock_search_results.return_value.results = []
         mock_search_results.return_value.total = 0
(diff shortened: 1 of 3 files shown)


anthonyhashemi:
I've just realised we have the NCN bug here since we have 
```
    @property
    def neutral_citation(self) -> str:
        """
        :return: The neutral citation of the search result
        """

        return self._get_xpath_match_string("search:extracted/uk:cite/text()")

```

on `SearchResult`.

We will need a `best_human_identifier`, but it gets complicated now since this is a search result and could be either judgemtn or press summary and eventually others.
First option is to just have something like

```
```
    @property
    def neutral_citation(self) -> str:
        """
        :return: The neutral citation of the search result
        """

        return self._get_xpath_match_string("search:extracted/uk:cite/text()") or self._get_xpath_match_string("search:extracted/PRESS_SUMMARY_NEUTRAL_CITATION_PATH")

```

thoughts @jacksonj04 @dragon-dxw ?

jacksonj04 (approved):
LGTM, we just need to yank the judgment-specific flavours from the API as well.

anthonyhashemi:
> LGTM, we just need to yank the judgment-specific flavours from the API as well.

Yeah I had added the judgment specific method to the api just so we didnt have to remember to filter by judgment in PUI and EUI but since it is only needed in PUI now we can probably remove it when we want but not an urgent thing

PR #1107, opened 2023-08-08
Push forwards node/debian

dragon-dxw:
`docker build .` works, now

(fab build invokes the copy of Dockerfile in compose; but that isn't used in prod)

Babel 9 requires Node 14 which isn't in Debian bullseye so we upgrade to bookworm, but bookworm changes how system users handle their home directories so we need to be explicit.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ ENV BUILD_ENV ${BUILD_ENVIRONMENT}
 WORKDIR ${APP_HOME}
 
 RUN addgroup --system django \
-    && adduser --system --ingroup django django --home /home/django
+    && adduser --system --ingroup django --home /home/django django
 
 
 # Install required system dependencies
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG PYTHON_VERSION=3.11-slim-bullseye
+ARG PYTHON_VERSION=3.11-slim-bookworm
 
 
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -53,7 +53,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   # cleaning up unused files
   && apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false \
   && rm -rf /var/lib/apt/lists/*
-RUN npm i -g sass
```
dragon-dxw, line 56: Might as well fail earlier in the deployment.

package-lock.json
```diff
@@ -11,7 +11,7 @@
       "devDependencies": {
         "@babel/core": "^7.17.2",
         "@babel/preset-env": "^7.16.11",
-        "babel-loader": "^8.2.5",
```
dragon-dxw, line 14: I regenerated package-lock but there are no changes from the reverted-revert.

Dockerfile
```diff
@@ -38,7 +38,7 @@ ENV BUILD_ENV ${BUILD_ENVIRONMENT}
 WORKDIR ${APP_HOME}
 
 RUN addgroup --system django \
-    && adduser --system --ingroup django django
+    && adduser --system --ingroup django django --home /home/django
```
rjw1, line 41: I personally wouldnt have a trailing option.

rjw1 (approved):
One minor quibble but otherwise looks good. (It just reads wrong in my head)

Dockerfile
jacksonj04, line 41:
```suggestion
    && adduser --system --ingroup django --home /home/django django
```

PR #1109, opened 2023-08-09
Improve handling of parser failure states

jacksonj04:
- Updates API Client to v14 and addresses breaking changes with metadata-setting operations
- Removes use of `is_failure` state flag
- Fixes review button not always being available
- Renders HTML where parsing has succeeded, but the document is at a failure URI because of an incorrect NCN
- Display of failure XML is now prettier, and has suitable coaching language to explain what you're actually looking at

### Before

editor staging caselaw nationalarchives gov uk_failures_TDR-2023-MQ6

### After

localhost_3000_failures_TDR-2023-MQ6

### DIFF
--- a/ds_caselaw_editor_ui/templates/judgment/full_text_html.html
+++ b/ds_caselaw_editor_ui/templates/judgment/full_text_html.html
@@ -6,21 +6,18 @@
     {% include "includes/judgment/metadata_panel_static.html" with return_to="html" %}
   {% endif %}
   {% include "includes/judgment/view_controls.html" with selected_tab=1 %}
-  {% if judgment_content %}
-    {% if judgment.is_failure %}
-      <div class="container">
-        <div class="page-notification--failure">This document is in a 'failure' state.</div>
-        <pre style="white-space:pre-wrap;
-                    background:#F6F6F6;
-                    border:1px solid #AAA;
-                    padding:1em">{{ judgment_content }}</pre>
-      </div>
-    {% else %}
-      {% autoescape off %}
-        {{ judgment_content }}
-      {% endautoescape %}
-    {% endif %}
+  {% if document.failed_to_parse %}
+    <div class="container">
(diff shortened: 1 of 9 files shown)


ds_caselaw_editor_ui/templates/judgment/full_text_html.html
```diff
@@ -8,9 +8,15 @@
   {% include "includes/judgment/view_controls.html" with selected_tab=1 %}
   {% if judgment_content %}
     {% if judgment.is_failure %}
-      <pre>
-        {{ judgment_content }}
-      </pre>
+      <div class="container">
+        <div class="page-notification--failure">This document is in a 'failure' state.</div>
+        <h2>Raw HTML</h2>
```
dragon-dxw, line 13: There's no guarantee it's HTML -- it could be a parser log. But this is still better.

dragon-dxw (approved):

ds_caselaw_editor_ui/templates/judgment/full_text_html.html
```diff
@@ -8,9 +8,15 @@
...
+        <p>This HTML is presented for debugging purposes.</p>
```
anthonyhashemi, line 14: if the html is fine, say if its a fialure just because the ncn wasnt found can we render the html?

judgments/views/document_full_text.py
```diff
@@ -17,14 +17,12 @@ def get_context_data(self, **kwargs):
 
         version_uri = self.request.GET.get("version_uri", None)
 
-        if not context["document"].is_editable:
-            context["judgment_content"] = context["document"].content_as_xml
-            context["metadata_name"] = context["document"].uri
+        if context["document"].failed_to_parse:
+            context["document_error_log"] = context["document"].content_as_xml()
```
anthonyhashemi, line 21: lovely

dragon-dxw (approved):

PR #1112, opened 2023-08-10
Reimplement the ability to delete documents.

jacksonj04:
**This PR needs known extra work to restrict scope of document deletion.**

Re-implement the ability for editors to delete documents. This is done via a link in the 'more drawer', which is only visible if the document is in a safe to delete state (currently the document is unpublished).

### DIFF
--- a/ds_caselaw_editor_ui/templates/judgment/delete.html
+++ b/ds_caselaw_editor_ui/templates/judgment/delete.html
@@ -1,0 +1,30 @@
+{% extends "layouts/judgment_with_sidebar.html" %}
+{% load i18n %}
+{% block judgment_header %}
+  {% include "includes/judgment/metadata_panel_static.html" with judgment=judgment %}
+{% endblock judgment_header %}
+{% block judgment_content %}
+  <h2>Delete this {{ document.document_noun }}</h2>
+  {% if document.safe_to_delete %}
+    <p>
+      If this {{ document.document_noun }} has previously been made public, you must follow the process for informing data reusers of a takedown.
+    </p>
+    <p>
+      <b>Deleting this {{ document.document_noun }} cannot be undone.</b>
+    </p>
+    <form aria-label="publish {{ document.document_noun }}"
+          action="{% url 'delete' %}"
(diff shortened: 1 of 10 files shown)


ds_caselaw_editor_ui/templates/judgment/delete.html
```diff
@@ -0,0 +1,30 @@
...
+  {% include "includes/judgment/metadata_panel_static.html" with judgment=judgment %}
+{% endblock judgment_header %}
+{% block judgment_content %}
+  <h2>Delete this {{ document.document_noun }}</h2>
+  {% if document.safe_to_delete %}
+    <p>
+      If this {{ document.document_noun }} has previously been made public, you must follow the process for informing data reusers of a takedown.
+    </p>
+    <p>
+      <b>Deleteing this {{ document.document_noun }} cannot be undone.</b>
```
dragon-dxw, line 13: typo: deleting

judgments/tests/test_document_delete.py
```diff
@@ -0,0 +1,85 @@
...
+from django.test import TestCase
+from django.urls import reverse
+from factories import JudgmentFactory
+
+
+class TestDocumentDelete(TestCase):
+    @patch("judgments.utils.view_helpers.get_document_by_uri_or_404")
+    @patch("judgments.utils.api_client.document_exists")
+    @patch("judgments.utils.api_client.get_document_type_from_uri")
+    def test_document_hold_view(self, document_type, document_exists, mock_judgment):
```
dragon-dxw, line 14: I don't think this function name encapsulates what the test is doing -- recommend rename or docstring.

dragon-dxw: "hold view has a button which links to the delete page?"

anthonyhashemi:
We ended up reinstating the delete button as it was before removal: https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/1124 to keep things familiar from editors as we have found that making UX changes need to be put through testing as we don't want to frustrate users with surprising changes.

PR #1120, opened 2023-08-16
Setup cypress e2e tests with simple login test test and configure CI to run the test suite against staging

anthonyhashemi:
## Changes in this PR:
- Setup cypress e2e tests 
- Wrote first simple cypress test: test login is successful
- Added GH action to run cypress tests against staging when:
     -  at 8am every day so we have a baseline of whether staging is good when we start in the morning.
     -  as a precondition step for the release flow, so that if the tests fail on staging when we have triggered a release, then it blocks the release/deployment.
     
## Verification of GH actions
I have verified the test runs and passes by adjusting the GH actiomn workflow file to run on push but I have now removed that. You can see the runs in the checks for this branch.

The only thing I havent verified is that deployment/release gets blocked if it fails when triggered by a release but I will just monitor that when I do a release this evening.

https://trello.com/c/VJEPR1yP/1287-run-e2e-cypress-test-suite-against-staging-on-every-successful-deployment-to-staging

### DIFF
--- a/.github/workflows/e2e_cypress_tests.yml
+++ b/.github/workflows/e2e_cypress_tests.yml
@@ -1,24 +1,20 @@
-name: E2E Tests on new deployments of staging
+name: E2E Cypress Tests
 
 on:
-  push:
-    # branches:
-    #   - main
+  workflow_call:
+    secrets:
+      CYPRESS_USERNAME:
+        required: true
+      CYPRESS_PASSWORD:
+        required: true
 
 jobs:
-  build-deploy-test:
+  run-e2e-cypress-test:
     runs-on: ubuntu-latest
 
     steps:
       - name: Checkout Code Repository
         uses: actions/checkout@v3
-
-      # - name: Wait for Deployment
-      #   run: |
-      #     # Poll AWS CodeDeploy status since dxws deployment of EUI
-      #     # to staging is currently triggered when we merge to main
-      #     # and they use AWS CodeDeploy to do it.
-      #     # Example: use AWS CLI to check deployment status
 
       - name: Install Node.js and npm
         uses: actions/setup-node@v2
--- a/.github/workflows/e2e_cypress_tests.yml
+++ b/.github/workflows/e2e_cypress_tests.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Install Node.js and npm
         uses: actions/setup-node@v2
         with:
-          node-version: 14
+          node-version-file: .node-version
 
       - name: Install Cypress
         run: npm install cypress
--- a/.github/workflows/e2e_cypress_tests.yml
+++ b/.github/workflows/e2e_cypress_tests.yml
@@ -2,6 +2,10 @@ name: E2E Cypress Tests
 
 on:
   workflow_call:
+    inputs:
+      CYPRESS_BASE_URL:
+        required: true
+        type: string
     secrets:
       CYPRESS_USERNAME:
         required: true
@@ -25,4 +29,4 @@ jobs:
         run: npm install cypress
 
       - name: Run Cypress Tests
-        run: CYPRESS_USERNAME=${{secrets.CYPRESS_USERNAME}} CYPRESS_PASSWORD=${{secrets.CYPRESS_PASSWORD}} npx cypress run --headless
+        run: CYPRESS_BASE_URL=${{inputs.CYPRESS_BASE_URL}} CYPRESS_USERNAME=${{secrets.CYPRESS_USERNAME}} CYPRESS_PASSWORD=${{secrets.CYPRESS_PASSWORD}} npx cypress run --headless
--- a/.github/workflows/e2e_staging_tests.yml
+++ b/.github/workflows/e2e_staging_tests.yml
@@ -1,7 +1,6 @@
 name: E2E Staging Tests
 
 on:
-  push:
   schedule:
     - cron: "0 8 * * *"
   workflow_call:
--- a/.github/workflows/e2e_staging_tests.yml
+++ b/.github/workflows/e2e_staging_tests.yml
@@ -4,9 +4,15 @@ on:
   push:
   schedule:
     - cron: "0 8 * * *"
+  workflow_call:
+    secrets:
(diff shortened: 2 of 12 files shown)


cypress/e2e/login.cy.js
```diff
@@ -0,0 +1,18 @@
+describe("Login Test", () => {
+  it("should successfully log in", () => {
+    // Visit the login page
+    cy.visit(
+      "https://editor.staging.caselaw.nationalarchives.gov.uk/accounts/login/",
```
anthonyhashemi, line 5: I want to adjust this to be use env vars to determine which deployment it should run against. You might want to run against your local!

.github/workflows/e2e_cypress_tests.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Code Repository
+        uses: actions/checkout@v3
+
+      - name: Install Node.js and npm
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
```
ahosgood, line 26:
```suggestion
          node-version-file: .node-version
```

ahosgood: This way the version is the same as your project, although I would recommend renaming `.node-version` to `.nvmrc` (https://github.com/nvm-sh/nvm/blob/master/README.md#nvmrc)

anthonyhashemi: ah great shout, cheers

.github/workflows/e2e_cypress_tests.yml
```diff
@@ -0,0 +1,32 @@
+name: E2E Cypress Tests
+
+on:
+  workflow_call:
```
dragon-dxw, line 4: ah, we've moved to explicitly triggered? Excellent.

cypress/e2e/login.cy.js
```diff
@@ -0,0 +1,16 @@
+describe("Login Test", () => {
```
dragon-dxw, line 1: It'd be nice to have a "login and then X" templated out for future users -- this might want to be a reusable function

.github/workflows/e2e_staging_tests.yml
```diff
@@ -0,0 +1,21 @@
+name: E2E Staging Tests
+
+on:
+  schedule:
+    - cron: "0 8 * * *"
```
dragon-dxw, line 5: where will this report to?

README.md
```diff
@@ -191,6 +191,25 @@ topic and `public-` and `private-asset-buckets` for S3.
...
+We have started adding cypress tests so we have some E2E tests to be happier with critical E2E workflows.
+These should be kept to a minimum since they are slow to run, resource intensive and harder to debug than smaller integration and unit tests.
+
+However, we should try to add all critical editor workflows here to give confidence we don't introduce regressions in new releases.
+
+NOTE: For now we should point the tests to staging, but eventually we may have a dedicated testing set up.
+
+To run them locally, you can run:
+
+`CYPRESS_baseUrl=base_eui_url CYPRESS_USERNAME=username CYPRESS_PASSWORD=password npx cypress run --headless`
```
dragon-dxw, line 205: `CYPRESS_BaseUrl` is capitalised everywhere else?

README.md
```diff
@@ -191,6 +191,25 @@ topic and `public-` and `private-asset-buckets` for S3.
...
+
+You can also run the tests interactively with:
+
+`CYPRESS_baseUrl=base_eui_url CYPRESS_USERNAME=username CYPRESS_PASSWORD=password npx cypress open`
```
dragon-dxw, line 209: as above: `CYPRESS_baseUrl` is capitalised elsewhere

dragon-dxw (approved):

README.md
anthonyhashemi: It has to be `CYPRESS_baseUrl` since the env var in cypress is `baseUrl`. 

`CYPRESS_` is the prefix for env vars used by cypress.

I could change `CYPRESS_USERNAME` to `CYPRESS_username`, 

or even all of them to `cypress_....`

Reference: https://docs.cypress.io/guides/guides/environment-variables

cypress/e2e/login.cy.js
anthonyhashemi: yup definitely, that is the plan. I was going to leave that for the first non login test.

.github/workflows/e2e_staging_tests.yml
anthonyhashemi: great question -it currently won't report anywhere... I assumed we had alerts set up for any failed builds on main but thats not true I dont think.

let me think about it

.github/workflows/e2e_cypress_tests.yml
anthonyhashemi: yeah I have changed it so that this workflow file doesnt trigger, but is a base for `.github/workflows/e2e_staging_tests.yml` which calls this general workflow with staging as the base url

.github/workflows/e2e_cypress_tests.yml
anthonyhashemi: I'm hesitant to change .node-version to .nvmrc mainly because our deployments are currently handled by dxw (contractors on our team)'s inhouse deployment setup(based off terraform) and so I dont have visibility on whether changing this could cause an issue.

@dragon-dxw do you know if that would be problematic?

But I will definitely make the suggested change of `node-version-file: .node-version`

README.md
anthonyhashemi: scratch that - I was able to use CYPRESS_BASE_URL and it worked, thanks for raising.

I had assumed it had to be CYPRESS_baseUrl just because thats what it is called inside cypress.

.github/workflows/e2e_staging_tests.yml
anthonyhashemi: I've created slack channel: #da-caselaw-staging-e2e-tests and subscribed to the "E2E Staging Tests" workflow using the github slack integration

PR #1122, opened 2023-08-17
re-order the static metadata panel

Donna-H:
## Changes in this PR:
re-order the static metadata panel to resemble order of metadata form
https://trello.com/c/vwukCv33/1248-eui-re-order-the-static-metadata-panel-to-match-the-form-metadata-panel
## Screenshots of UI changes:
Metadata form order 
form

### Before

Judgment
before-static

Press Summary
before-press-summary

### After
Judgment
after-judgment

Press Summary
press-summary-after


Terry-Price (changes requested):
@Donna-H 
All good I like the order and the fuller title for 'type' – only thing is can we alter the metadata for Press Summary, so it renders with a capital 'S' on 'Summary'?

dragon-dxw (approved):

PR #1123, opened 2023-08-17
Reinstate delete button for failures

anthonyhashemi:
## Changes in this PR:
In https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/1103/files we removed the delete button that was accessible on "failure" URI documents when we shouldn't have as it was still in use.

This adds it back. Only difference to the original is that it now just takes you to the home page/unpublished list when you confirm deletion with the green message banner telling you it has been deleted as opposed to previously where there was an interim "delete.html" page which said the same thing and had a link to take you back to the homepage.

There was a different design where deleting a document would not have this big red button, it would be in the more toolbar and take you to a new page to delete but we decided against changing that just yet so that things dont change too much for the editors. Reference for that work here: https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/1112/files

https://trello.com/c/FIuUf94n/1246-eui-reinstate-the-delete-button

### Before
Screenshot 2023-08-17 at 17 07 46

### After
Screenshot 2023-08-17 at 17 13 17

### DIFF
--- a/ds_caselaw_editor_ui/templates/includes/judgment/toolbar.html
+++ b/ds_caselaw_editor_ui/templates/includes/judgment/toolbar.html
@@ -44,6 +44,16 @@
           <path d="M233.4 406.6c12.5 12.5 32.8 12.5 45.3 0l192-192c12.5-12.5 12.5-32.8 0-45.3s-32.8-12.5-45.3 0L256 338.7 86.6 169.4c-12.5-12.5-32.8-12.5-45.3 0s-12.5 32.8 0 45.3l192 192z" />
         </svg>
       </a>
+      {% if judgment.is_failure %}
+        <form action="{% url 'delete' %}" method="post">
+          {% csrf_token %}
+          <input type="hidden" name="judgment_uri" value="{{ judgment.uri }}" />
+          <input type="submit"
+                 name="assign"
+                 class="button-secondary judgment-toolbar__delete"
+                 value="Delete" />
+        </form>
+      {% endif %}
     </div>
     <div class="judgment-toolbar__more">{% include "includes/judgment/toolbar_more.html" %}</div>
     {% if version %}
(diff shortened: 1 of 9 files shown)


ds_caselaw_editor_ui/templates/includes/judgment/toolbar.html
```diff
@@ -44,6 +44,16 @@
           <path d="M233.4 406.6c12.5 12.5 32.8 12.5 45.3 0l192-192c12.5-12.5 12.5-32.8 0-45.3s-32.8-12.5-45.3 0L256 338.7 86.6 169.4c-12.5-12.5-32.8-12.5-45.3 0s-12.5 32.8 0 45.3l192 192z" />
         </svg>
       </a>
+      {% if judgment.is_failure %}
+        <form action="{% url 'delete' %}" method="post">
+          {% csrf_token %}
+          <input type="hidden" name="judgment_uri" value="{{ judgment.uri }}" />
+          <input type="submit"
+                 name="assign"
+                 class="button-secondary judgment-toolbar__delete"
```
anthonyhashemi, line 53: @Terry-Price I now remmeber you asked to change the colour to `color__failure`

I will need to update  `&__delete {` in `.judgment-toolbar`
in `ds_caselaw_editor_ui/sass/includes/_judgment_text_toolbar.scss` to do that

anthonyhashemi:
> I don't like how we now have a second row of buttons at a reasonable screenwidth now, but it's not the worst thing and reinstating is important.

@dragon-dxw this is the same layout we had before and I wanted to go with what we had before so editors aren't confused when the button was reinstated. I do agree with you that we can improve this, 100%. Hopefully we can do some UX around this soon

ds_caselaw_editor_ui/templates/includes/judgment/toolbar.html
anthonyhashemi: spoke with Terry on slack. Gone with color_failure

PR #1135, opened 2023-08-24
Configure default date time formats

dragon-dxw:
We use "7 Jan 1999" for the date

### DIFF
--- a/judgments/views/judgment_edit.py
+++ b/judgments/views/judgment_edit.py
@@ -1,3 +1,5 @@
+import datetime
+
 from caselawclient.Client import MarklogicAPIError, api_client
 from django.contrib import messages
 from django.http import HttpResponseRedirect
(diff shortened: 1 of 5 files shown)


dragon-dxw:
[The publication date is in ISO format not English.](https://trello.com/c/T0ZSgyUu/1271-the-publication-date-is-in-iso-format-not-english)

dragon-dxw:
Check that the date editing field isn't in the new format!

ds_caselaw_editor_ui/sass/includes/_judgment_text.scss
```diff
@@ -0,0 +1,393 @@
+.judgment {
```
anthonyhashemi, line 1: I think this was mitakenly comitted. I imagine you had this in unstaged files since we no longer copy it over from PUI / care about having a local copy of it and so stopped git ignoring it in this repo. Can just uncommit and delete the file.

judgments/views/judgment_edit.py
```diff
@@ -41,7 +43,21 @@ def post(self, request, *args, **kwargs):
 
             # Date
             new_date = request.POST["judgment_date"]
-            api_client.set_judgment_date(judgment_uri, new_date)
+            try:
+                new_date_as_date = datetime.datetime.strptime(new_date, r"%d %b %Y")
```
anthonyhashemi, line 47: Does this mean all dates in the db from when this is merged in will look different to those from before?

If so, do we plan to run a migration on the previous dates?

Or is this just accounting for the fact that the judgment_date will now display in the new presentation format and so we need to revert back to the old format when writing to the db?

dragon-dxw: No migrations -- everything's ISO8601 under the hood. This is just presentation converting back.

dragon-dxw: Might want to hint at right format / support multiple formats. Django looks like it has a way to do it, but couldn't get it working.

dragon-dxw:
This might be deprecated in Django 5 

  /usr/local/lib/python3.11/site-packages/django/conf/__init__.py:267: RemovedInDjango50Warning: The USE_L10N setting is deprecated. Starting with Django 5.0, localized formatting of data will always be enabled. For example Django will display numbers and dates using the format of the current locale.
    warnings.warn(USE_L10N_DEPRECATED_MSG, RemovedInDjango50Warning)

jacksonj04 (approved):

PR #1138, opened 2023-08-30
Upgrade to Cypress 13

dragon-dxw:
This upgrades node's requests library to suppress a vulnerability warning. (Since Cypress isn't run in production, there isn't a lot to worry about).

This might update the version of node required to Node 16 from reading the generated package.json, but .nodeversion is 18 already.


dragon-dxw:
the login test runs OK locally.

anthonyhashemi (approved):

PR #1143, opened 2023-09-04
⚠️ Create own MarklogicApiClient instance and set user agent

dragon-dxw:
Check that this is working correctly on staging, since it changes how we handle the environment variables and is challenging to test.

### DIFF
--- a/judgments/tests/utils/test_utils.py
+++ b/judgments/tests/utils/test_utils.py
@@ -9,6 +9,7 @@ from django.test import TestCase
 from factories import UserFactory
 
 import judgments
+from judgments.utils import api_client as api_client_real
 from judgments.utils import (
     editors_dict,
     ensure_local_referer_url,
@@ -311,3 +312,8 @@ class TestEditorsDict:
         assert editors_dict() == [
             {"name": "active_editor", "print_name": "active_editor"},
         ]
+
+
+class TestApiClient:
+    def test_user_agent(self):
+        assert "ds-caselaw-editor" in api_client_real.session.headers["User-Agent"]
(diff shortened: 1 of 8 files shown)


dragon-dxw:
[Add user-agent to ingester, etc](https://trello.com/c/CzpD33r0/1320-add-user-agent-to-ingester-etc)

judgments/utils/__init__.py
```diff
@@ -6,14 +6,26 @@
...
 from django.contrib.auth.models import Group, User
 
 from .aws import copy_assets
 
+api_client = MarklogicApiClient(
+    host=settings.MARKLOGIC_HOST,
+    username=settings.MARKLOGIC_USER,
+    password=settings.MARKLOGIC_PASSWORD,
+    use_https=settings.MARKLOGIC_USE_HTTPS,
+    user_agent=f"ds-caselaw-editor/unknown {DEFAULT_USER_AGENT}",
```
anthonyhashemi, line 26: Is this because we made it so we expect this to look like "app/version"?

judgments/tests/utils/test_utils.py
```diff
@@ -9,6 +9,7 @@
 from factories import UserFactory
 
 import judgments
+from judgments.utils import api_client as api_client_real
```
anthonyhashemi, line 12: why does it need to be renamed in this namespace?

judgments/utils/__init__.py
dragon-dxw: yep; we don't currently record the version of editor/public-ui inside the code.

judgments/tests/utils/test_utils.py
dragon-dxw: It didn't; changed.

anthonyhashemi (approved):

PR #1148, opened 2023-09-06
(FCL-61) Ruff support

dragon-dxw:
Use Ruff as a linter with most of the options turned on.


dragon-dxw:
For now, consider using `stages: manual` to not have this be part of the standard run?

anthonyhashemi (approved):

PR #1151, opened 2023-09-07
QOL improvement - better template debugging.

timcowlishaw:
## Changes in this PR:
add TEMPLATE_DEBUG=1 to .env to clearly see where undefined variable errors are raised in templates.


config/settings/local.py
```diff
@@ -9,6 +9,11 @@
 # https://docs.djangoproject.com/en/dev/ref/settings/#allowed-hosts
 ALLOWED_HOSTS = ["*"]
 
+TEMPLATE_DEBUG = env("TEMPLATE_DEBUG", default=False) == 1
+
+if TEMPLATE_DEBUG:
```
jacksonj04, line 14: You could squish this down to

```suggestion
if env("TEMPLATE_DEBUG", default=False) == 1:
```

if you want and it might be marginally more readable than the `... = ... == ...` form above, but no strong feelings.

jacksonj04 (approved):
Two very weakly felt comments, but LGTM 📦

config/settings/local.py
anthonyhashemi: what is the expected truthy response, is it not True?

PR #1167, opened 2023-09-20
first iteration of "add a note" UI

timcowlishaw:
## Changes in this PR:

This adds, a (currently non-functional) 'add a note' field on the confirmation page for the publish, unpublish, put on hold and take off hold actions, behind a feature flag, so that Rose can test it! I've modified the designs slightly based on the earlier wireframes to accomodate their new position on the confirmation page, and make the copy and style consistent with GDS guidelines on optional form field (as well our own form field styles elsewhere).

**After deploying, the feature flag `notes_on_state_change` will need to be added in the django admin UI, at which point the feature can be enabled in labs.**

https://trello.com/c/Hvh7udC2/1355-eui-add-a-note-component-mvp-for-testing

## Screenshots of UI changes:
Screenshot 2023-09-20 at 16 26 27

### DIFF
--- a/ds_caselaw_editor_ui/templates/includes/note-form.html
+++ b/ds_caselaw_editor_ui/templates/includes/note-form.html
@@ -1,11 +1,8 @@
 <div class="note-form">
-  <h2>
-    <label class="note-form__label" for="write-a-note">Write a note</label>
-  </h2>
+  <label class="note-form__label" for="write-a-note">Add a note for the editoral team (optional)</label>
   <textarea class="note-form__text-area"
             id="write-a-note"
             name="write-a-note"
             rows="8"
-            placeholder="Write your message..."
             aria-describedby="write-a-note"></textarea>
 </div>
--- a/ds_caselaw_editor_ui/sass/includes/_note.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_note.scss
@@ -1,14 +1,20 @@
 .note-form {
   &__label {
-    margin-bottom: 1rem;
+    display: block;
+    margin-bottom: calc($spacer__unit / 4);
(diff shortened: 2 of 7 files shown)


timcowlishaw:
Tiny change from the screenshot, after a chat with Gaz, the label now reads "Add a note for the editorial team (optional)"

timcowlishaw:
[[EUI] 'ADD A NOTE' component MVP - for testing](https://trello.com/c/Hvh7udC2/1355-eui-add-a-note-component-mvp-for-testing)

ds_caselaw_editor_ui/templates/includes/note-form.html
```diff
@@ -1,11 +1,8 @@
 <div class="note-form">
-  <h2>
-    <label class="note-form__label" for="write-a-note">Write a note</label>
-  </h2>
+  <label class="note-form__label" for="write-a-note">Add a note for the editoral team (optional)</label>
```
dragon-dxw, line 2: feels slightly weird being called note-form when it's not actually a whole form.

Should this actually be a complete form that redirects back to/doesn't leave the current page, so that individual forms for many actions don't need to be note-aware?

dragon-dxw (approved):

ds_caselaw_editor_ui/templates/includes/note-form.html
timcowlishaw: Aha yep, that definitely needs reconsidering - that component was actually originally made ages ago on the style guide page, so we should update to make sure its name reflects its actual role! If it's ok with you i'll merge as is, then do a little tidy up when i'm back in on tuesday, that way Rose can test with what's there!

PR #1171, opened 2023-09-21
Check that users are editors or devs before deleting

dragon-dxw:
Delete button should always appear in More for editors and superusers
/delete action should fail for non-editor, non-superusers

https://linear.app/tna/issue/FCL-102/make-delete-button-available-for-all-editors

### DIFF
--- a/ds_caselaw_editor_ui/templates/includes/judgment/toolbar_more.html
+++ b/ds_caselaw_editor_ui/templates/includes/judgment/toolbar_more.html
@@ -1,4 +1,4 @@
-{% load i18n %}
+{% load i18n user_permissions %}
 <hr>
 <div class="judgment-toolbar__more__menu">
   <div class="judgment-toolbar__more__menu__group">
@@ -46,9 +46,9 @@
       </li>
     </ul>
   </div>
-  {% if user.is_superuser %}
+  {% if user|is_superuser_or_editor %}
     <div class="judgment-toolbar__more__menu__group">
-      <h4>Superuser tools</h4>
+      <h4>Editor tools</h4>
       <form action="{% url 'delete' %}" method="post">
         {% csrf_token %}
         <input type="hidden" name="judgment_uri" value="{{ judgment.uri }}" />
--- a/judgments/tests/test_document_delete.py
+++ b/judgments/tests/test_document_delete.py
@@ -1,12 +1,21 @@
 from unittest.mock import patch
 
-from django.contrib.auth.models import User
+from django.contrib.auth.models import Group, User
 from django.test import TestCase
 from django.urls import reverse
 from factories import JudgmentFactory
 
 
 class TestDocumentDelete(TestCase):
+    def setUp(self):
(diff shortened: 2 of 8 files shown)


ds_caselaw_editor_ui/templates/includes/judgment/toolbar_more.html
```diff
@@ -46,7 +46,7 @@ <h4>Downloads</h4>
       </li>
     </ul>
   </div>
-  {% if user.is_superuser %}
+  {% if user|is_superuser_or_editor %}
     <div class="judgment-toolbar__more__menu__group">
       <h4>Superuser tools</h4>
```
jacksonj04, line 51: This probably wants changing as well!

judgments/tests/test_document_delete.py
```diff
@@ -32,6 +43,32 @@ def test_document_delete_flow_if_safe(self, mock_document, mock_invalidate_cache
         mock_document.return_value.delete.assert_called_once()
         mock_invalidate_caches.assert_called_once()
 
+    @patch("judgments.views.delete.invalidate_caches")
+    @patch("judgments.views.delete.get_document_by_uri_or_404")
+    def test_document_delete_flow_if_not_editor(
+        self, mock_document, mock_invalidate_caches
+    ):
+        document = JudgmentFactory.build(
```
anthonyhashemi, line 51: minor readability suggestion:
```
        deletable_document = JudgmentFactory.build(
            uri="deltest/4321/123",
            safe_to_delete=True,
        )
        mock_document.return_value = deletable_document
```

needs `safe_to_delete` to be pulled into the `JudgmentFactory` to be consistent with how we treat other properties.

Ideally we wouldn't need to add every new property manually to the Factory and instead just base the factory of the class - could be made easier by just using https://factoryboy.readthedocs.io/en/stable/examples.html. Also pulling the factory into the apiclient would make sense then we can utilise the same factory here and PUI.

judgments/tests/test_document_toolbar.py
```diff
@@ -2,14 +2,52 @@
...
+        editor_group.save()
+        self.editor_user = User.objects.get_or_create(username="ed")[0]
+        self.editor_user.groups.add(editor_group)
+        self.editor_user.save()
+        self.standard_user = User.objects.get_or_create(username="alice")[0]
+        self.super_user = User.objects.create_superuser(username="clark")
+
+    @patch("judgments.utils.view_helpers.get_document_by_uri_or_404")
+    @patch("judgments.utils.api_client.document_exists")
+    def test_superuser_tools_if_editor(self, document_exists, mock_judgment):
```
anthonyhashemi, line 25: similar to nick's comment, the fact that we still refer to these tools as superuser tools is a bit misleading since they can be accessed by superusers and editors, so worth updating the test names.

Maybe privilleged_tools or anything other than superuser really

anthonyhashemi (approved):
Approving as code looks good but left couple maintainability suggestions

PR #1173, opened 2023-09-22
add type column to version history

Donna-H:
## Changes in this PR:
Add type column to version history
https://trello.com/c/jyTZaNMi/1359-eui-version-history-add-column-to-table-to-indicate-type-of-version
### Before
before-judgment
before-press-summary

### After
after-judgment
after-press-summary

### DIFF
--- a/ds_caselaw_editor_ui/templates/judgment/history.html
+++ b/ds_caselaw_editor_ui/templates/judgment/history.html
@@ -12,6 +12,7 @@
           <th scope="col">Version</th>
           <th scope="col">Date submitted</th>
           <th scope="col">Time</th>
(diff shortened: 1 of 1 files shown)


dragon-dxw (approved):
As much as I want to say this should be constant across all the documents, I bet it won't be :)

This should probably be the first bit of data to get changed into a tooltip or fold out when we run out of space in the table :)

Donna-H:
[[EUI] VERSION HISTORY Add column to table to indicate type of version](https://trello.com/c/jyTZaNMi/1359-eui-version-history-add-column-to-table-to-indicate-type-of-version)

PR #1178, opened 2023-09-26
Remove 'add a note' from labs

timcowlishaw:
## Changes in this PR:

it's useful to keep the UI behind an invisible flag for testing purposes, but given it's not feature-complete, we don't want editors to be able to turn it on independently just yet.


timcowlishaw:
[Remove 'add a note' flag from labs](https://trello.com/c/3Ovog996/1376-remove-add-a-note-flag-from-labs)

anthonyhashemi (approved):

PR #1179, opened 2023-09-27
Show version type in version table

timcowlishaw:
## Changes in this PR:

Previously the document history table showed the document type (ie Judgment or Press Summary) when what we wanted to show was the type of _version_ (ie submission or enrichment). This PR fixes, that, pending a new release of the API client (with [this currently open branch merged](https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/402)

https://trello.com/c/jyTZaNMi/1359-eui-version-history-add-column-to-table-to-indicate-type-of-version

### Before
Screenshot 2023-09-27 at 14 15 41

### After
Screenshot 2023-09-27 at 14 15 11

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_typography.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_typography.scss
@@ -1,21 +1,3 @@
-a {
-  color: $color__link-blue;
-  text-decoration: underline;
-
-  &:hover {
-    color: $color__link-blue-hover;
-    text-decoration: none;
-  }
-
-  &:visited {
-    color: $color__link-blue-visited;
-  }
-
-  &:active {
-    color: $color__link-blue-active;
-  }
-
-  &:focus {
-    @include focus-default;
-  }
+.muted {
+  color: $color__dark-grey;
 }
--- a/ds_caselaw_editor_ui/sass/includes/_typography.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_typography.scss
@@ -1,3 +1,25 @@
+a {
+  color: $color__link-blue;
+  text-decoration: underline;
+
+  &:hover {
+    color: $color__link-blue-hover;
+    text-decoration: none;
+  }
+
+  &:visited {
+    color: $color__link-blue-visited;
+  }
+
+  &:active {
+    color: $color__link-blue-active;
+  }
+
+  &:focus {
+    @include focus-default;
+  }
+}
+
 .muted {
   color: $color__dark-grey;
 }
--- a/ds_caselaw_editor_ui/templates/judgment/history.html
+++ b/ds_caselaw_editor_ui/templates/judgment/history.html
@@ -10,9 +10,8 @@
       <thead>
         <tr>
           <th scope="col">Version</th>
-          <th scope="col">Date submitted</th>
-          <th scope="col">Time</th>
-          <th scope="col">Type</th>
+          <th scope="col">Date created</th>
+          <th scope="col">Details</th>
         </tr>
       </thead>
       <tbody>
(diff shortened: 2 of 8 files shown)


timcowlishaw:
[[EUI] VERSION HISTORY Add column to table to indicate type of version](https://trello.com/c/jyTZaNMi/1359-eui-version-history-add-column-to-table-to-indicate-type-of-version)

ds_caselaw_editor_ui/templates/judgment/history.html
```diff
@@ -23,7 +23,7 @@ <h2>Document version history</h2>
             </td>
             <td>{{ version.get_latest_manifestation_datetime|display_datetime_date }}</td>
             <td>{{ version.get_latest_manifestation_datetime|display_datetime_time }}</td>
-            <td>{{ judgment.document_noun|title }}</td>
+            <td>{{ judgment.get_latest_manifestation_type|display_friendly_version_type }}</td>
```
anthonyhashemi, line 26: is it worth adding a test to show this view returns the right html snippet given this change?

timcowlishaw: aha yep good idea, done!

timcowlishaw:
Screenshot 2023-10-03 at 14 05 07

Folks little update to use the new API version and the annotations functionality therein. We'll need to do some further work elsewhere to make sure those annotations contain useful info, but i think it's a much better approach!

Thanks

ds_caselaw_editor_ui/sass/includes/_document_version_history.scss
```diff
@@ -4,4 +4,8 @@
     width: auto;
     border-collapse: collapse;
   }
+
+  &__unknown-label {
```
jacksonj04, line 8: I wonder if it's worth taking a leaf from Bootstrap et-al and making this something like `.muted` in the typography classes?

judgments/templatetags/document_utils.py
```diff
@@ -11,10 +13,16 @@ def get_title_to_display_in_html(document_title, document_type):
 
 
 @register.filter
-def display_datetime_date(datetime):
-    return datetime.strftime("%d %b %Y")
+def display_datetime_date(dt):
+    if isinstance(dt, datetime.datetime):
```
jacksonj04, line 17: This feels smelly to me; is there anywhere calling this function where it actually expects to receive a string?

timcowlishaw: Aha tidied! that was my mistake - it *can* receive a None, but not a string.

judgments/templatetags/document_utils.py
```diff
@@ -11,10 +11,8 @@ def get_title_to_display_in_html(document_title, document_type):
...
 @register.filter
-def display_datetime_date(datetime):
-    return datetime.strftime("%d %b %Y")
-
-
-@register.filter
-def display_datetime_time(datetime):
-    return datetime.strftime("%H:%M")
+def display_datetime(dt):
+    if dt is not None:
```
jacksonj04, line 15: I'm still not totally happy with this; if there's a state where we're trying to display a version without a datetime then perhaps blowing up is the correct outcome?

timcowlishaw: aha right you are, fixed!

timcowlishaw:
@jacksonj04 this should now be ready for version 16.0.0!

judgments/tests/factories.py
```diff
@@ -66,7 +66,8 @@ def build(cls, **kwargs) -> Judgment:
         judgment = judgment_mock()
         version = judgment.copy()
         version.version_number = 1
-        version.get_latest_manifestation_datetime = datetime.datetime(2023, 9, 26, 12)
+        version.version_created_datetime = datetime.datetime(2023, 9, 26, 12)
+        version.annotation = '{"type": "submission"}'
```
jacksonj04, line 70: I wonder if it's worth using `VersionAnnotation()` here to build a 'real' annotation string, that way it's a bit more resilient against any future changes we might make.

jacksonj04 (approved):
LGTM! Once we get more structured data in there we can start to make this history richer, but for now this is great stuff 👍🏻 

A couple of tiny comments but nothing I feel strongly about.

ds_caselaw_editor_ui/sass/includes/_typography.scss
```diff
@@ -1,21 +1,3 @@
-a {
```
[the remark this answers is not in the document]
timcowlishaw: oh good spot, that should *not* have happened!

PR #1182, opened 2023-09-28
Remove unused aws functions

dragon-dxw:
A lot of the aws functions have been migrated to the custom-api-client and invoked there rather than in the editor.

There's probably also useful work in migrating the still-used functions to the client, but didn't fancy taking that step right now.

Please double check that the functions deleted are in fact not used anywhere when reviewing :)


anthonyhashemi (approved):
🔥

PR #1222, opened 2023-10-17
Disable delete button where document is not deletable

jacksonj04:
## Changes in this PR:

The delete button is now disabled for documents which don't report themselves as being deletable (currently documents which are published). The CSS for this button didn't make this state apparent, so this now makes the button match other `danger` styling and introduces a new `button-small` class which trims down the padding etc.

### Before

localhost_3000_ukut_aac_2023_205 (1)

### After

localhost_3000_ukut_aac_2023_205


dragon-dxw:
Probably needs a design pass since it doesn't look very disabled.

dragon-dxw (approved):

PR #1223, opened 2023-10-18
Further desaturate disabled danger button

jacksonj04:
## Changes in this PR:

Unify the 'disabled' states of all button types, so it's obvious that they're not for clicking.

### Before

localhost_3000_ukut_tcc_2023_61 (1)

### After
localhost_3000_ewhc_admin_2023_1249_

## Future work

We should migrate all the button styles to the shared frontend and tweak them so we're not having to use `!important`.

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_buttons.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_buttons.scss
@@ -6,21 +6,27 @@
   }
 }
 
+@mixin button-disabled {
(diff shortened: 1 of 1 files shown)


timcowlishaw (approved):
I would be tempted to just use the darker grey we use elsewhere, but looks good to me!

gazaston:
Agree that we should use a 'standard' grey — or we run the risk of having various styles of disabled buttons.

PR #1236, opened 2023-10-25
Add more useful links to right of navbar

jacksonj04:
## Changes in this PR:

In readiness for making stats easy to download, we need to be able to link to the functionality. The closest thing we have to an abstract top-level link is the link to Labs, so this PR repurposes that area in the right of the navbar into a more general-purpose set of links/information, including:

- The name of the currently signed-in user
- (If the user is a superuser) a link to the admin area
- A link to Labs
- A link to sign the current user out

### Before

localhost_3000_ (1)

### After

localhost_3000_


timcowlishaw (approved):
aha, great stuff. All good to go in once the breadcrumbs yak is freshly shaved :-)

PR #1248, opened 2023-10-31
Add ability to download summary statistics

jacksonj04:
This is behind the `stats` feature flag.


judgments/tests/test_stats.py
```diff
@@ -0,0 +1,35 @@
...
+        ]
+
+        response = self.client.get(reverse("stats_combined_csv"))
+
+        decoded_response = response.content.decode("utf-8")
+
+        assert response.status_code == 200
+        assert (
+            decoded_response
+            == "Column 1,Column 2,Column 3,Column 4\r\nValue 1,Value 2,,7\r\n"
```
dragon-dxw, line 34: Does this escape quotes, etc?

judgments/views/stats.py
```diff
@@ -0,0 +1,44 @@
...
+        headers={
+            "Content-Disposition": 'attachment; filename="tna-fcl-combined-stats-{datetime}.csv"'.format(
+                datetime=datetime.now(pytz.timezone("Europe/London")).strftime(
+                    "%Y%m%d%H%M%S",
+                ),
+            ),
+        },
+    )
+
+    writer = csv.writer(response)
```
dragon-dxw, line 39: https://docs.python.org/3/library/csv.html#id3
> If `newline=''` is not specified, newlines embedded inside quoted fields will not be interpreted correctly, and on platforms that use `\r\n` linendings on write an extra `\r` will be added. It should always be safe to specify `newline=''`, since the csv module does its own ([universal](https://docs.python.org/3/glossary.html#term-universal-newlines)) newline handling.

dragon-dxw: still haunted by [python2 csv](https://python.readthedocs.io/en/v2.7.2/library/csv.html) not being unicode compliant, but pretty sure they've fixed that from reading the docs.

docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
       - postgres
     volumes:
       - .:/app:z
+      - ../ds-caselaw-custom-api-client:/apiclient
```
dragon-dxw, line 13: This feels out of place -- is this test code?

dragon-dxw (approved):

PR #1260, opened 2023-11-07
User-specific links in top menu now no longer show if logged out

jacksonj04:
These appear on the login page, and won't work properly.

### DIFF
--- a/ds_caselaw_editor_ui/templates/layouts/base.html
+++ b/ds_caselaw_editor_ui/templates/layouts/base.html
@@ -42,27 +42,33 @@
       <div class="page-header__navbar page-header__navbar-right">
         <nav>
           <ol>
-            <li>
-              {% if user.first_name %}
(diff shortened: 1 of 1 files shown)


ds_caselaw_editor_ui/templates/layouts/base.html
```diff
@@ -42,29 +42,33 @@
...
           <ol>
-            <li>
-              {% if user.first_name %}
-                {{ user.first_name }}
-              {% else %}
-                {{ user.username }}
-              {% endif %}
-            </li>
-            {% if user|is_superuser %}
+            {% if user.is_authenticated %}
```
dragon-dxw, line 45: can the user ever not be authenticated?

dragon-dxw: (Yes, the login page)

dragon-dxw (approved):

PR #1267, opened 2023-11-09
Use rich enrichment annotations to display them nicely in the timeline

jacksonj04:
Since we [updated the Privileged API to use new Annotations](https://github.com/nationalarchives/ds-caselaw-privileged-api/pull/254), we can now properly detect when a new version has arrived as the result of enrichment.

This PR neatly surfaces that fact in the Timeline view.


timcowlishaw (approved):
super!

PR #1284, opened 2023-11-14
Add 'submission-oriented' history to Labs

jacksonj04:
Implement a first pass at the new 'submission-oriented' history view and make it available in Labs.

This new view introduces a hierarchy of versions, with actual submissions by editors taking precedence. It has some known missing features, to be implemented progressively as we gain a better understanding of requirements. In particular, it only properly handles submission and enrichment structured events; edit events and state changes are not yet supported (and we lack the data to do so anyway).

A large chunk of this is trying to handle the many possible cases and interactions between our various levels of annotation richness (legacy, type-only and fully payload), as well as variation in the level of detail in a payload.


ds_caselaw_editor_ui/templates/includes/judgment/history/document_data.html
```diff
@@ -0,0 +1,28 @@
+{% load i18n %}
+<div class="document-history__data-panel my-1">
+  <ul class="document-history__data-panel__details">
+    <li>
+      <span class="judgment-metadata-panel__key">{% translate "judgments.ncn" %}</span>
```
dragon-dxw, line 5: I thought we'd decided to junk new translations

judgments/views/document_history.py
```diff
@@ -1,5 +1,117 @@
...
+        sorted_ml_versions = sorted(
+            context["document"].versions_as_documents,
+            key=lambda x: x.version_number,
+        )
+
+        # Start from the point of there being no valid submission in progress
+        submission = None
+
+        # Event sequence number increments on each new event
+        event_sequence_number = 0
```
dragon-dxw: I'm not 100% sure that event numbers aren't going to change... but should be fine.

judgments/views/document_history.py
```diff
@@ -5,10 +5,109 @@
...
+
+        try:
+            structured_data = json.loads(version.annotation)
+
+            # If we're here, we can unpack the structured data and that means we can figure out what next.
+
+            if structured_data["type"] == VersionType.SUBMISSION.value:
+                # This version is a submission
+
+                # If a submission is already open, pop it on the stack.
```
dragon-dxw, line 96: add "and start a new submission" to comment?

judgments/views/document_history.py
```diff
@@ -5,10 +5,109 @@
...
+                    "document": version,
+                    "events": [],
+                    "sequence_number": submission_sequence_number,
+                }
+
+            elif not submission:
+                # This isn't a submission, it's an event causing a new version. There's also a possibility that the
+                # first structured event we see happens outside a submission, in which case it gets its own special
+                # orphan submission.
+                submission = {
```
dragon-dxw: comment doesn't seem to agree with code: "first structured event we see... in which case it gets [special treatment]" but there's no conditionals in this area.

judgments/views/document_history.py
```diff
@@ -5,10 +5,109 @@
...
+
+    # Submission sequence only gets incremented for new submissions
+    submission_sequence_number: int = 0
+
+    # These are all MarkLogic versions, not necessarily our concept of submissions... so unpack it!
+    for version in history:
+        # Iterate this at the beginning of the loop because 1-index is nicer for humans
+        event_sequence_number += 1
+
+        try:
```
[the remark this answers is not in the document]
jacksonj04: I definitely agree, it's already been unpacked once which is where the event generator comes from! I'll have a ponder and see if there are any more slices, but it might just be that this really is a big procedural bit of suck.

judgments/views/document_history.py
jacksonj04: This is strictly "the order this event happened in the history stack as presented to the user" rather than any kind of idempotent identifier. We can revisit it once we start merging in things like publish events alongside versions to see if there's a better way, this is mostly just so the JavaScript has something to reference for expanding events.

ds_caselaw_editor_ui/templates/includes/judgment/history/document_data.html
jacksonj04: This was a copy/paste job I left intact so all the metadata boxes behave the same, but I believe they're on the slate for a revamp (or at least a unification since at the moment they have three different styles) so hopefully this goes away soon.

judgments/views/document_history.py
jacksonj04: The "is this the first structured event" logic is implicit here; if we've already had a structured event then all code paths mean `submission` has a value, ie it's only ever possible to reach this point if what the comment says holds true. I used to have an `else` with a nested `if`, but ruff complained that an `elif` was cleaner.

Agree this comment (even if technically right) sucks at explaining what is going on though.

judgments/views/document_history.py
```diff
@@ -21,21 +21,19 @@ class EventDict(TypedDict):
 
 
 class SubmissionDict(TypedDict):
-    pass
+    events: NotRequired[list[EventDict]]
```
dragon-dxw, line 24: TIL about NotRequired

judgments/views/document_history.py
```diff
@@ -138,29 +128,49 @@ def structure_document_history_by_submissions(
...
+) -> EventDict:
+    event_object: EventDict = {
+        "data": event_data,
+        "datetime": event_document.version_created_datetime,
+        "document": event_document,
+        "sequence_number": event_sequence_number,
+        "marklogic_version": event_document.version_number,
+    }
+
+    if "payload" in event_data and "submitter" in event_data["payload"]:
```
dragon-dxw, line 169: possibly better as: `if "submitter" in event_data.get("payload",{}):`?

dragon-dxw (approved):

dragon-dxw:
```
         {% if event.data.calling_agent %}
            {{ event.data.calling_agent }}
          {% else %}
            Unknown agent
          {% endif %}
```

might be better phrased as `{{event.data.calling_agent|default="Unknown agent"}}` (untested)

judgments/views/document_history.py
jacksonj04: It's tighter, I'm not sure it's more readable.

PR #1285, opened 2023-11-15
Fix TDR spacing issue in metadata panel

timcowlishaw:
## Changes in this PR:

This was a tricky one to track down but not to fix - when the TDR has four characters in the last position, and one of them is em-width (an 'm' or 'w'), at a very specific screensize (~1400px) it pops underneath the status badge and is illegible. 
([Here's an example](https://editor.caselaw.nationalarchives.gov.uk/ukut/lc/2023/271) in production that will replicate). The fix is to tweak the breakpoint at which the width of that metadata panel becomes constrained with a max-width, to give it a little more room. This is an exception to our standard breakpoints, but i think a good one to make - the current extra-large breakpoint is 1200px which would exacerbate the problem if we were using it here.

I've also changed the label for the document type from 'Doc type' to 'Type', as discussed on slack - we don't specify 'document' in any other labels here, and it's already unambiguous.

https://trello.com/c/j2eb1fCH/1470-eui-tdr-spacing-issue

### Before
Screenshot 2023-11-15 at 11 24 00

### After
Screenshot 2023-11-15 at 11 29 02

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_judgment_view_controls.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_judgment_view_controls.scss
@@ -1,5 +1,5 @@
 .judgment-view-controls {
-  @media (min-width: 1400px) {
+  @media (min-width: 1500px) {
     width: 67.5%;
     margin: 0 auto;
   }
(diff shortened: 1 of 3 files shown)


timcowlishaw:
[[EUI] TDR spacing issue](https://trello.com/c/j2eb1fCH/1470-eui-tdr-spacing-issue)

jacksonj04 (approved):

PR #1291, opened 2023-11-17
Refactor and extend test factories for documents

jacksonj04:
**This should be merged after https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/1290**

To support robust testing of version history behaviour, refactor `JudgmentFactory` into a more generic `DocumentFactory`, and a new `DocumentVersionFactory` which can handle version-specific functionality.

### DIFF
--- a/judgments/views/document_full_text.py
+++ b/judgments/views/document_full_text.py
@@ -35,7 +35,9 @@ class DocumentReviewPDFView(DocumentView):
         context = super().get_context_data(**kwargs)
 
         if not context["document"].pdf_url:
-            msg = f"Document \"{context['document'].name}\" does not have a PDF."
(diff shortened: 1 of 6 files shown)


judgments/views/document_full_text.py
```diff
@@ -35,7 +35,9 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
         if not context["document"].pdf_url:
-            msg = f"Document \"{context['document'].name}\" does not have a PDF."
+            msg = 'Document "{document_name}" does not have a PDF.'.format(
```
dragon-dxw, line 38: I'd have probably just used an f-string here

dragon-dxw (approved):

PR #1296, opened 2023-11-21
Feature/handle orphan submissions

jacksonj04:
Orphaned submissions are those which cannot be definitively attached to a previous submission, due to a break in structured history since (or the entire absence of) a previous structured submission.


dragon-dxw (approved):
Commit seems good, but I think it's worth having a chat to bring me up to speed on what an orphaned submission is.

PR #1310, opened 2023-11-28
Enriched status

dragon-dxw:
Needs some minor formatting and layout tweaks, but adds time since enrichment to the detail edit page (six months ago/never etc). Mouse over it for an exact timestamp.


dragon-dxw:
This was a proof of concept, we've got a better approach coming now.

PR #1361, opened 2024-01-05
Add a view which shows documents pending enrichment

jacksonj04:


judgments/views/reports.py
```diff
@@ -0,0 +1,31 @@
+from django.views.generic import TemplateView
+
+from judgments.utils import api_client
+
+
+class Index(TemplateView):
+    template_name = "reports/index.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
```
dragon-dxw, line 10: We should probably have a very basic test for this.

dragon-dxw (approved):

PR #1365, opened 2024-01-09
Add a management command to re-enrich items in the queue

jacksonj04:


judgments/management/commands/enrich_next_in_reenrichment_queue.py
```diff
@@ -0,0 +1,23 @@
+from django.core.management.base import BaseCommand
+
+from judgments.utils import api_client
+
+TARGET_ENRICHMENT_VERSION = 6
```
dragon-dxw, line 5: We want to change this later to reference the db, but this is MVP

dragon-dxw (approved):

PR #1376, opened 2024-01-22
Reparse button

dragon-dxw:
Hidden behind 'reparse' feature flag. Requires `REPARSE_SNS_TOPIC` environment variable setting.


jacksonj04 (approved):
LGTM. Before we switch on for general users we need to make sure that the `.reparse()` method is actually sending the right metadata payload!

PR #1377, opened 2024-01-24
Add report on documents needing re-parsing

jacksonj04:
As a precursor to automatic re-parsing, we have added the necessary tooling to MarkLogic and API Client to report on documents which haven't been parsed with the latest version.

This adds a report to EUI which exposes those documents.

### DIFF
--- a/ds_caselaw_editor_ui/templates/reports/awaiting_enrichment.html
+++ b/ds_caselaw_editor_ui/templates/reports/awaiting_enrichment.html
@@ -4,7 +4,7 @@
   <div class="standard-text-template">
     <h1>Documents awaiting enrichment</h1>
     <p>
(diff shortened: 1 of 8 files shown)


ds_caselaw_editor_ui/templates/reports/awaiting_enrichment.html
```diff
@@ -4,7 +4,7 @@
   <div class="standard-text-template">
     <h1>Documents awaiting enrichment</h1>
     <p>
-      These documents have not yet been enriched with the target version of the enrichment engine, version {{ target_enrichment_version }}.
+      These documents have not yet been enriched with the latest version of the enrichment engine, version <b>{{ target_enrichment_version }}</b>.
```
dragon-dxw, line 7: Is it worth documenting here that ones that have been tried recently aren't here?

dragon-dxw (approved):

PR #1397, opened 2024-02-08
Watch JS and assets while running the server

pezholio:
When getting set up I was confused about the lack of assets. I then realised I forgot to build the assets, which is documented a bit further down.

To make `fab run` more of a one-shot process (and to save us having to spawn multiple terminal windows), I’ve added a couple of steps that:

- Run `npm install`; and
- Watch the Sass and JS for changes

The `npm` tasks are backgrounded and the log output is in seperate files. Ideally I’d want to have this logged to stdout along with the Django logs, but I couldn’t see an easy way to do this in a fabfile.

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -34,6 +34,11 @@ def container_exec(cmd, container_name="django", check_returncode=False):
     return result
 
 
+def background_exec(cmd, logfile):
+    "Run a command in the background and capture logs."
+    local(f"nohup {cmd} &> {logfile}.log &")
+
+
 def postgres_exec(cmd, check_returncode=False):
     "Execute something in the 'postgres' Docker container."
     return container_exec(cmd, "postgres", check_returncode)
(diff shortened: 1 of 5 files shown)


pezholio:
Yeah, no probs 👍

fabfile.py
```diff
@@ -68,9 +73,20 @@ def start(c, container_name=None):
...
+    Install NPM packages
+    """
+    local("npm install")
+
+
 @task
 def run(c):
     start(c, "django")
+    npm_install(c)
+    background_exec("npm run start-sass", "assets")
```
dragon-dxw, line 88: How does this differ from `npm run build`? (I have no idea.)

fabfile.py
```diff
@@ -68,9 +73,20 @@ def start(c, container_name=None):
     local(cmd)
 
 
+@task
+def npm_install(c):
+    """
+    Install NPM packages
+    """
+    local("npm install")
```
dragon-dxw, line 81: Is this quick the second time?

dragon-dxw (approved):

pezholio:
@timcowlishaw I guess you could do that, but I wonder if it makes sense to have `pip install` as part of the build too in that case? I'm coming back into Python after a few years, so there may be some subtleties missing

fabfile.py
pezholio: `build` just builds the assets, while `start-sass` watches them for changes. That said, we probably need to run `collectstatic` again to pick up those changes

fabfile.py
pezholio: Yeah, it's a couple of seconds max

timcowlishaw:
@pezholio aha i wouldn't worry too much - from memory i think pip install already runs on both steps? I'm not certain though. Definitely not a blocker in any case!

pezholio:
I've updated this to use `nodemon` to watch the files for changes and run `collectstatic`, so now watching the assets should work as expected.

fabfile.py
pezholio: @dragon-dxw I've tweaked it so `watch` works as expected and runs `collectstatic` on each change 👍

dragon-dxw (approved):
The changes look good -- I like it not reloading all the python dependencies *again* :)

PR #1400, opened 2024-02-09
Do not enrich if enriched recently

dragon-dxw:
Some of the problems recently encountered might have been caused by two enrichment processess occurring simultanously. Avoid triggering enrichment f it has been enriched recently.


pezholio (approved):
LGTM - I do think it might be worth revisiting the whole way we approach enrichment, but this solves the initial problem that users are getting as I understand it 👍

PR #1408, opened 2024-03-01
Update spacing between buttons

Donna-H:
## Changes in this PR:
Update spacing between buttons  in 'more drawer'
https://trello.com/c/Q5oDasDR/1638-eui-fix-vertical-spacing-in-more-drawer-buttons

### Before
before-1
2-before

### After
after-2

after-1

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_buttons.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_buttons.scss
@@ -3,6 +3,7 @@
     padding: 0.4rem 0.5rem !important;
     margin-top: 0.5em !important;
     font-size: 0.8rem !important;
+    margin-bottom: 0.5rem !important;
   }
 }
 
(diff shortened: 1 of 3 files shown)


Donna-H:
[EUI: Fix vertical spacing in 'more drawer' buttons](https://trello.com/c/Q5oDasDR/1638-eui-fix-vertical-spacing-in-more-drawer-buttons)

CristinaRO (approved):

dragon-dxw:
Things that are adjacent to this PR worth thinking about but not blockers:
* that border around the box is very big
* we're adding another button soon which might make that button area quite large
* is this caused by the frontend at all? will it interfere with that work?

dragon-dxw (approved):

Donna-H:
> Things that are adjacent to this PR worth thinking about but not blockers:
> 
> * that border around the box is very big
> * we're adding another button soon which might make that button area quite large
> * is this caused by the frontend at all? will it interfere with that work?

@dragon-dxw Okay, Heather suggested this can be brought up at Roadmap day @timcowlishaw will be joining online .

PR #1411, opened 2024-03-05
Do not enrich if enriched recently -- disable button if can't enrich/reparse

dragon-dxw:


CristinaRO (approved):
It makes sense to me. One comment to be addressed or ignored.

dragon-dxw:
Probably archaic.

PR #1414, opened 2024-03-13
Disable enrichment queue

dragon-dxw:
Disabled this function in order to make sure that documents were not being sent to enrichment (and therefore locked) before editors had a chance to edit them. We intend to revisit this decision later so keeping the code intact makes sense.


adam-murray (approved):
Looks good to me!

PR #1490, opened 2024-06-25
[FCL-128] Remove unused templates and CSS

adam-murray:
## Changes in this PR:
* Reviewed all templates and removed any that are not in use
* Removed associated css

https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-128

### After


ds_caselaw_editor_ui/sass/includes/_browse_by_court.scss
```diff
@@ -1,61 +0,0 @@
-.judgment-browse-by-court {
```
jacksonj04, line 1: Do we also want to remove the unused `templates/includes/browse_by_court.html`?

jacksonj04 (changes requested):

dragon-dxw (approved):

jacksonj04 (approved):

PR #1491, opened 2024-06-26
Ensure we send at least one document to be reparsed

dragon-dxw:
It's likely for a while that the document chosen to be reparsed has no docx. Skip it and try the next.

### DIFF
--- a/judgments/management/commands/reparse_next_in_reparse_queue.py
+++ b/judgments/management/commands/reparse_next_in_reparse_queue.py
@@ -18,7 +18,8 @@ class Command(BaseCommand):
             ),
         )
 
-        for document_details in document_details_to_parse[:NUMBER_TO_PARSE]:
+        counter = 0
+        for document_details in document_details_to_parse:
             document_uri = document_details[0]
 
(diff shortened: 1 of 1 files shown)


judgments/management/commands/reparse_next_in_reparse_queue.py
```diff
@@ -18,13 +18,18 @@ def handle(self, *args, **options):
...
+        counter = 0
+        for document_details in document_details_to_parse:
             document_uri = document_details[0]
 
             document = api_client.get_document_by_uri(document_uri.replace(".xml", ""))
 
             self.stdout.write(f"Attempting to reparse document {document.name}...")
             if document.reparse():
                 self.stdout.write("Reparse request sent.")
+                counter = counter + 1
```
jacksonj04, line 30:
```suggestion
                counter += 1
```

jacksonj04 (changes requested):
One small syntactic suggestion.

CodeClimate is also waving flags about complexity. It might be worth decomposing this so that the actual "get a document, try reparse it, return `True`/`False`" is in a separate method.

jacksonj04 (approved):

PR #1508, opened 2024-07-03
Fcl 149/refactor spacing

jlhdxw:
## Changes in this PR:

Includes:

 - [x] Add $space-1 - $space-12 variables
 - [x] Update existing CSS $spacer-unit calculations to have static variables
 - [x] Add font size variables
 - [x] Add line height variables
 - [x] Update all font sizes to use variables
 - [x] Update all line heights to use variables

https://national-archives.atlassian.net/browse/FCL-149

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_login.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_login.scss
@@ -4,13 +4,13 @@
   padding-bottom: $space-8;
 
   @media (min-width: $grid-breakpoint-small) {
-    padding-top: calc($space-4 * 5);
-    padding-bottom: calc($space-4 * 5);
+    padding-top: $space-20;
+    padding-bottom: $space-20;
   }
 
   @media (min-width: $grid-breakpoint-large) {
-    padding-top: calc($space-4 * 8);
-    padding-bottom: calc($space-4 * 8);
+    padding-top: $space-32;
+    padding-bottom: $space-32;
   }
 }
 
--- a/ds_caselaw_editor_ui/sass/includes/_login.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_login.scss
@@ -1,16 +1,16 @@
 .login-background {
   background-color: $color-light-grey;
-  padding-top: $spacer-unit * 2;
-  padding-bottom: $spacer-unit * 2;
+  padding-top: $space-8;
+  padding-bottom: $space-8;
 
   @media (min-width: $grid-breakpoint-small) {
-    padding-top: $spacer-unit * 5;
-    padding-bottom: $spacer-unit * 5;
+    padding-top: calc($space-4 * 5);
+    padding-bottom: calc($space-4 * 5);
   }
 
   @media (min-width: $grid-breakpoint-large) {
-    padding-top: $spacer-unit * 8;
-    padding-bottom: $spacer-unit * 8;
+    padding-top: calc($space-4 * 8);
+    padding-bottom: calc($space-4 * 8);
   }
 }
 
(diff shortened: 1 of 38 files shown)


ds_caselaw_editor_ui/sass/includes/_login.scss
```diff
@@ -1,16 +1,16 @@
...
   background-color: $color-light-grey;
-  padding-top: $spacer-unit * 2;
-  padding-bottom: $spacer-unit * 2;
+  padding-top: $space-8;
+  padding-bottom: $space-8;
 
   @media (min-width: $grid-breakpoint-small) {
-    padding-top: $spacer-unit * 5;
-    padding-bottom: $spacer-unit * 5;
+    padding-top: calc($space-4 * 5);
```
dragon-dxw, line 7: should this be a $space-20? See also the four lines below.

ds_caselaw_editor_ui/sass/includes/_metadata.scss
```diff
@@ -28,35 +28,36 @@
     align-items: start;
     > div {
       @media (min-width: $grid-breakpoint-large) {
-        height: calc(1.8 * $spacer-unit);
+        height: 1.8rem;
```
dragon-dxw, line 31: Interesting that this and the next isn't $space derived...

ds_caselaw_editor_ui/sass/includes/_standard-text-template.scss
```diff
@@ -1,7 +1,7 @@
 .standard-text-template {
   @include container;
-  margin-top: calc($spacer-unit * 5);
-  margin-bottom: calc($spacer-unit * 8);
+  margin-top: calc($space-4 * 5);
```
dragon-dxw, line 3: These calculations make me suspicious that "space-4" is actually the important unit here; since you're writing this instead of calc($space-1 * 20);

dragon-dxw (approved):

ds_caselaw_editor_ui/sass/includes/_login.scss
jlhdxw: It's a good observation. I decided to intentional only go up to $space-16 as realistically anything beyond that suggests we've got some odd/very specific styling going on, and either this is a unique case and doesn't really fit into our normal spacing, or it could be solved in a different way.

With that said, one of the reasons for doing this work was to reduce the cognitive load reading CSS, so maybe we just increase the available spaces and set this to it.

ds_caselaw_editor_ui/sass/includes/_metadata.scss
jlhdxw: Height/width are dimensions rather than spacing around/inside something, so they shouldn't be defined by our spacing definitions.

ds_caselaw_editor_ui/sass/includes/_standard-text-template.scss
jlhdxw: I think `calc($space-1 * 20) == 5rem` is more confusing than `calc($space-4 * 5) == 5rem` as while working on this we will know `$space-4 is 1rem` as it's the most common spacing.

If we defined them as `$space-1_4` (as in that equals 0.25, or quarter of our base unit) it ties that variable name to that value. If we decided to adjust all our spacing (to make everything bigger or smaller) the variable naming being tied to their values means they make less sense now.

Similar to how we have a variable called $font-roboto - if we decide to change that font everywhere, we'd have to change the variable everywhere, making the whole exercise of using variables redundant. We'd be better off calling that $font-display or something similar depending on how it's used.

ds_caselaw_editor_ui/sass/includes/_login.scss
jlhdxw: Have added the extra spacing to do away with the calcs 👍

PR #1544, opened 2024-07-25
Fcl 160/extract judgment toolbar

jlhdxw:
## Changes in this PR:

 - Refactor toolbar
 - Add to styleguide

https://national-archives.atlassian.net/browse/FCL-160

image

### DIFF
--- a/ds_caselaw_editor_ui/templates/includes/judgment/toolbar_more.html
+++ b/ds_caselaw_editor_ui/templates/includes/judgment/toolbar_more.html
@@ -1,78 +1,80 @@
-{% load i18n user_permissions waffle_tags %}
-<hr>
-<div class="judgment-toolbar__more__menu">
-  <div class="judgment-toolbar__more__menu__group">
-    <h4>Submitter</h4>
-    <p>
-      <a href="mailto:{{ document.source_email }}">{{ document.source_name }}</a>
-    </p>
-  </div>
-  <div class="judgment-toolbar__more__menu__group">
-    <h4>Tools</h4>
-    <ul>
-      <li>
-        <a href="{{ jira_create_link }}"
-           target="_blank"
-           rel="noopener noreferrer">{% translate "judgment.create_in_jira" %}</a>
-      </li>
-      <li>
-        <a href="{% url 'document-history' document.uri %}">Document version history</a>
-      </li>
-      <li>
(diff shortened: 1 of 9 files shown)


Donna-H (approved):
All looks fine to me @jlhdxw

locale/en_GB/LC_MESSAGES/django.po
```diff
@@ -97,54 +97,6 @@ msgstr "No one"
 msgid "judgment.assign_cta"
 msgstr "Assign"
 
-#: ds_caselaw_editor_ui/templates/includes/judgment/toolbar.html
```
anthonyhashemi, line 100: Is this change because we are moving away from using django translations bit by bit? I think I remember that decision being made when I was still working on fcl but just wanted to double check

jlhdxw: Yep that's it :)

judgments/views/style_guide.py
```diff
@@ -10,6 +10,7 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         judgment = SearchResultFactory.build()
         context["judgment"] = judgment
+        context["document"] = judgment
```
anthonyhashemi, line 13: this is because we're moving towards using document as a generic object as opposed to judgment so that it can handle press summaries etc too, right?

ds_caselaw_editor_ui/templates/includes/judgment/toolbar_more.html
```diff
@@ -1,78 +1,80 @@
...
-           target="_blank"
-           rel="noopener noreferrer">{% translate "judgment.report_blocking_problem" %}</a>
-      </li>
-    </ul>
-  </div>
-  <div class="judgment-toolbar__more__menu__group">
-    <h4>Downloads</h4>
-    <ul>
-      {% if document.docx_url %}
+{% load user_permissions waffle_tags %}
```
anthonyhashemi, line 1: this diff is a bit funky in github, whats the main change Im looking for here?

judgments/views/style_guide.py
jlhdxw: I'm not aware of a decision to move towards using "document" over judgment, but that does make sense. If that is the case I wish I had known that sooner so I could've done that refactor as I've been doing other refactoring. 

I'll leave this like this for now and clarify with Dragon/Nick when they're back how they'd like to approach this. I'm happy to refactor as I go.

ds_caselaw_editor_ui/templates/includes/judgment/toolbar_more.html
jlhdxw: There isn't any change here other than moving the containing div into here rather than outside the include.

dragon-dxw (approved):

PR #1592, opened 2024-08-28
NCN mismatch fix

dragon-dxw:
## Changes in this PR
Decouple changing the NCN from changing the URI and allow editors to be explicit about when they are wanting to move.

## Jira card
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136

## Screenshots of UI changes:
image

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_notification_message.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_notification_message.scss
@@ -22,6 +22,11 @@
       color: $color-dark-blue;
     }
   }
+
+  &--warning {
+    @include notification;
+    background-color: #ff7900;
+  }
 }
 
 .inline-notification {
--- a/ds_caselaw_editor_ui/templates/includes/judgment/judgment_metadata_form.html
+++ b/ds_caselaw_editor_ui/templates/includes/judgment/judgment_metadata_form.html
@@ -68,7 +68,8 @@
         </div>
         <div class="metadata-component__actions">
           <input type="hidden" name="judgment_uri" value="{{ document_uri }}" />
-          <input type="submit" value="Save" />
+          <input type="submit" name="save_only" value="Save" />
+          <input type="submit" name="save_and_move" value="Save and Move" />
         </div>
       </div>
     </form>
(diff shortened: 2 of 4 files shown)


ds_caselaw_editor_ui/sass/includes/_notification_message.scss
```diff
@@ -22,6 +22,11 @@
       color: $color-dark-blue;
     }
   }
+
+  &--warning {
```
dragon-dxw, line 26: This should probably be using a colour from the style guide, but none feel semantically correct.

ds_caselaw_editor_ui/templates/includes/judgment/judgment_metadata_form.html
```diff
@@ -68,7 +68,8 @@
         </div>
         <div class="metadata-component__actions">
           <input type="hidden" name="judgment_uri" value="{{ document_uri }}" />
-          <input type="submit" value="Save" />
+          <input type="submit" name="save_only" value="Save" />
```
dragon-dxw, line 71: I'd like something cleverer here, perhaps, like a dropdown for Save and Move. But not sure it's worth it.

judgments/views/judgment_edit.py
```diff
@@ -64,14 +69,26 @@ def post(self, request, *args, **kwargs):
...
 
-            # If judgment_uri is a `failure` URI, amend it to match new neutral citation and redirect
-            if "failures" in judgment_uri and new_citation is not None:
-                new_judgment_uri = update_document_uri(judgment_uri, new_citation)
-                return redirect(
-                    reverse("edit-document", kwargs={"document_uri": new_judgment_uri}),
-                )
-
-            messages.success(request, "Document successfully updated")
+            invalidate_caches(judgment.uri)
```
dragon-dxw, line 72: Move invalidate caches earlier so we always do it.

judgments/views/judgment_edit.py
```diff
@@ -64,14 +69,26 @@ def post(self, request, *args, **kwargs):
...
+
+            if new_uri and new_uri.strip("/") != judgment_uri:
```
dragon-dxw, line 74: There's an ugly mismatch in whether there's a leading `/` in the uri.

dragon-dxw:
We didn't like the UI approach so changed to https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/1605

PR #1605, opened 2024-09-06
Fcl 257 banner bad uri

dragon-dxw:
We had a think about how we were doing https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/1592 and changed it so the UI only existed if there was a mismatch at all.

https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136

image

### DIFF
--- a/ds_caselaw_editor_ui/sass/includes/_notification_message.scss
+++ b/ds_caselaw_editor_ui/sass/includes/_notification_message.scss
@@ -22,6 +22,15 @@
       color: $color-dark-blue;
     }
   }
+
+  &--warning {
+    @include notification;
+    background-color: $color-yellow;
+    color: $color-almost-black;
+    a {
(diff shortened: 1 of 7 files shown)


ds_caselaw_editor_ui/sass/includes/_notification_message.scss
```diff
@@ -22,6 +22,15 @@
       color: $color-dark-blue;
     }
   }
+
+  &--warning {
+    @include notification;
+    background-color: #ff7700;
```
jlhdxw, line 28: Shall we use $color-yellow here? it looks basically the same as this and we'll be keeping things consistent:

image

jlhdxw (approved):
1 minor comment on the color, but like I said in the standup I'll be refactoring this component anyway so feel free to leave as is

PR #1635, opened 2024-09-23
[FCL-315] Do not provide mismatch banner for press summaries

dragon-dxw:
https://dxw.slack.com/archives/C04FQ3GFGUQ/p1727098797043679

previous behaviour:
image

## Jira

FCL-315


jacksonj04 (approved):
I don't like it, but it works until we properly retool press summaries.

PR #1646, opened 2024-09-25
FCL-300 | update navigation behaviour to hide unusable buttons

jlhdxw:
## Changes in this PR:

When a nav item isn't usable in judgment state, don't show the nav item.

Added some tests to reflect what the states should show.

The templatetag determines whether the navigation item has a link or not based off the document attributes, and the template decides whether to show the navigation item based on that. If we decide to change our mind, it'll be easy to switch this around in the template rather than changing all the logic.

https://national-archives.atlassian.net/browse/FCL-300

## Screenshots of UI changes:

image

image

### DIFF
--- a/judgments/templatetags/navigation_tags.py
+++ b/judgments/templatetags/navigation_tags.py
@@ -5,47 +5,38 @@ register = template.Library()
 
 
 def get_document_url(view, document):
-    return reverse(view, kwargs={"document_uri": document.uri})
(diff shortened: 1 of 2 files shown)


jlhdxw:
Closing this PR as discussion has been made whereby we don't want to hide navigation items: https://national-archives.atlassian.net/browse/FCL-300

PR #1647, opened 2024-09-25
FCL-336 | add extra cases for status colors

jlhdxw:
## Changes in this PR:

Fix for statuses that don't match the judgment format (from the editor metadata)

Separate question on why they don't follow the same format? Couldn't figure out why this might be. Ideally they would be consistent.

https://national-archives.atlassian.net/browse/FCL-336

### Before

image

### After

image

### DIFF
--- a/judgments/tests/test_templatetags.py
+++ b/judgments/tests/test_templatetags.py
@@ -16,11 +16,17 @@ class TestStatusTagColour:
     def test_colour_in_progress(self):
         assert status_tag(DOCUMENT_STATUS_IN_PROGRESS) == "in-progress"
 
+    def test_colour_in_progress_editor_status(self):
+        assert status_tag("In Progress") == "in-progress"
+
     def test_colour_published(self):
         assert status_tag(DOCUMENT_STATUS_PUBLISHED) == "published"
 
     def test_colour_hold(self):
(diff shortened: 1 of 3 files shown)


dragon-dxw (approved):

judgments/tests/test_templatetags.py
```diff
@@ -16,12 +16,18 @@ class TestStatusTagColour:
     def test_colour_in_progress(self):
         assert status_tag(DOCUMENT_STATUS_IN_PROGRESS) == "in-progress"
 
+    def test_colour_in_progress_editor_status(self):
+        assert status_tag("In Progress") == "in-progress"
```
jacksonj04, line 20:
```suggestion
        assert status_tag(DOCUMENT_STATUS_IN_PROGRESS) == "in-progress"
```

jacksonj04 (changes requested):
Super quick tweak to use a constant

judgments/tests/test_templatetags.py
jlhdxw: Ah sorry - if you check above I'm already doing it. This is testing when it's not the constant (i.e. the state coming from the search rather than the document)

jacksonj04 (approved):

PR #1650, opened 2024-09-26
FCL-324 | update downloads screen

jlhdxw:
## Changes in this PR:

Changing the text and removing the list from the downloads page

https://national-archives.atlassian.net/browse/FCL-324

### Before

image

### After

image


jacksonj04 (changes requested):
This feels like the changes to the static panel should really be their own PR, but all the content of the changes looks good!

jacksonj04 (approved):
🚀

PR #1651, opened 2024-09-26
FCL-341 | add associated documents nav item for judgments with associ…

jlhdxw:
## Changes in this PR:

Add "associated documents" to the judgment nav when they have associated documents and link to a page that lists the associated documents.

https://national-archives.atlassian.net/browse/FCL-341

## Screenshots of UI changes:

image


jacksonj04 (changes requested):
Changes LGTM - I'd personally prefer "related documents" than "associated documents" but that's just me 😄 

The only change I'd request is that the "remove sidebar" be sliced into its own PR so we can reason about the two changes separately. Also I love PRs which are nothing but removing old code.

jacksonj04 (approved):
:shipit:

PR #1653, opened 2024-09-27
FCL-316 | update judgment toolbar spacing

jlhdxw:
## Changes in this PR:

Change judgment toolbar spacing

https://national-archives.atlassian.net/browse/FCL-316

## Screenshots of UI changes:

image


jacksonj04 (approved):
LGTM; have we checked this behaves well on mobile? Not that I expect editors to be doing it from their phone, but best practice and all...

jlhdxw:
@jacksonj04 I've checked it on mobile and the whole EUI is basically unusable on mobile at this point. I think when they do the reimagining of the EUI work, whether it should be supported on mobile should probably play into that conversation. Ideally it would be mobile first to save us having to really think about it on mobile.

PR #1710, opened 2024-10-24
[FCL-400] Swap to using new in-API XSLT transformation

jacksonj04:
This now leans on using the XSLT for body transformation which exists within the API, rather than making a call out to MarkLogic.

## Jira

FCL-400

### DIFF
--- a/ds_caselaw_editor_ui/templates/judgment/full_text_html.html
+++ b/ds_caselaw_editor_ui/templates/judgment/full_text_html.html
@@ -28,7 +28,7 @@
            class="document-navigation-links">
         <a class="up" href="#start-of-document">Back to start</a>
       </div>
-      {{ document_html_content }}
+      {{ document_html }}
       {% include "includes/judgment_end_document_marker.html" %}
       <div id="end-of-document">
       {% endautoescape %}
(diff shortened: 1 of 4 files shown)


ds_caselaw_editor_ui/templates/judgment/full_text_html.html
```diff
@@ -28,7 +28,7 @@ <h2>Error log</h2>
            class="document-navigation-links">
         <a class="up" href="#start-of-document">Back to start</a>
       </div>
-      {{ document_html_content }}
```
dragon-dxw, line 31: Does `document_body` need to explcitly be a context? or should this be a `document.body`? (we already have `document.body` on line 9)

dragon-dxw: Ah, I see. It's the actual body we're rendering, as opposed to derived variables from the unmodified body. I think that's a potentially confusing name -- "body_to_render" or something?

dragon-dxw: ... but I can't see anything else that's using that PR yet. Confused as to why document_body exists.

jacksonj04: Fixed so the flow is clearer.

dragon-dxw (approved):

PR #1735, opened 2024-11-01
[FCL-443] Use signed image urls for saxon HTML transform

dragon-dxw:
## Jira

FCL-443


judgments/tests/fixtures/sample_judgment.xml
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jacksonj04, line 1: I wonder if it's worth building a library of these sample docs in the API Client test fixtures, which we can import (similar to the factories) for test purposes.

jacksonj04 (approved):

PR #1741, opened 2024-11-04
Reenable enrichment

dragon-dxw:

https://national-archives.atlassian.net/browse/FCL-138

### DIFF
--- a/judgments/management/commands/enrich_next_in_reenrichment_queue.py
+++ b/judgments/management/commands/enrich_next_in_reenrichment_queue.py
@@ -1,23 +1,17 @@
+import time
+
 from django.core.management.base import BaseCommand
 
 from judgments.utils import api_client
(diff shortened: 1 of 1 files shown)


jacksonj04 (commented):
Have we adjusted the queue query so it only plucks published judgments?

dragon-dxw:
Here's a pull request for that:
https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/755

But I'm not sure we want to restrict bulk enrichment it to just published documents -- the rationale behind that was that we didn't want enrichment locking things and interferring with editors ability to get their job done.

https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/716 and [TODO: priv API part of this] change how we lock things so that they'll automatically unlock after a period of time.

jacksonj04 (approved):

PR #1754, opened 2024-11-14
[FCL-466] Bump API client to v28

jacksonj04:
Give us the added confidence of the strict type checking when initialising a `Document`.

## Jira

FCL-466


dragon-dxw:
Please review the new commit I added before merging!

PR #1763, opened 2024-11-20
Use client document factories

dragon-dxw:
### After


judgments/tests/factories.py
```diff
@@ -23,70 +23,25 @@ class Meta:
...
-
 class DocumentVersionFactory(DocumentFactory):
     DocumentClass: TypeAlias = Document
 
     @classmethod
-    def build(cls, **kwargs) -> DocumentClass:
+    def build(
+        cls,
+        uri="test/2023/123",
+        html="<p>This is a Document Version.</p>",
```
jacksonj04, line 33: This is going to go away soon (since we deprecated `content_as_html`, it belongs in the `body`), but for now it may as well stay and we can deal with it when we actually deprecate things.

jacksonj04 (approved):

PR #1773, opened 2024-11-27
[FCL-493] Fix NCN/URI mismatch detector incorrectly raising false positives

jacksonj04:
Now we have fully standardised our approach to document URI strings our NCN/URI mismatch detector is incorrectly applying a leading forward slash during comparisons; stop doing this.

There is a change to the existing test for the banner being correctly shown so that both the positive and negative case are testing for (or against) the same thing using the same method, reducing the chances of accidentally introducing a false pass for the negative case.

FCL-493


jlhdxw (approved):
LGTM

PR #1790, opened 2024-12-05
Add more headers to the access log

rjw1:
This adds header which give us more information about the request. This is
useful for debugging and monitoring.
It will give us the viewer's IP address, the CloudFront request ID, and the
trace ID from ALB if present.


rjw1:
amz and amzn are typos on my part. its almost as if different amazon teams dont talk to each other.

dragon-dxw (approved):

PR #1843, opened 2025-01-15
Fcl 538/colour refactoring

jlhdxw:
## Changes in this PR:

Use TNA colours in EUI

https://national-archives.atlassian.net/browse/FCL-538


ds_caselaw_editor_ui/sass/includes/_basic_search_form.scss
```diff
@@ -40,7 +40,7 @@
   }
 
   &__search-term-input {
-    @include text_field;
+    @include text-field;
```
jacksonj04, line 43: I suspect this would be a _massive pain_ to fix, so if it is don't bother, but could we do this 'auto-fix all the underscores and dashes' in its own commit?

Terry-Price (approved):
Hi James,
It's quite difficult to really check this due to the variable names, but  trust you and happy to sign off. We can alway check the colours when they appear in the UI.

PR #1865, opened 2025-01-23
[FCL-625] Update Jira link generator with better identifier handling

jacksonj04:
This puts the best human identifier (where present) in the Jira summary field, and expands the description to include all identifiers which are known.

### DIFF
--- a/judgments/utils/link_generators.py
+++ b/judgments/utils/link_generators.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 from urllib.parse import quote, urlencode
 
-from caselawclient.models.identifiers.neutral_citation import NeutralCitationNumber
 from django.conf import settings
 from django.template import loader
 from django.urls import reverse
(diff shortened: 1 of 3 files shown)


judgments/utils/link_generators.py
```diff
@@ -80,31 +79,35 @@ def build_raise_issue_email_link(
...
+        summary_string = f"{best_human_identifier.value} / {document.body.name} / {document.consignment_reference}"
+    else:
+        summary_string = f"{document.body.name} / {document.consignment_reference}"
 
     editor_html_url = request.build_absolute_uri(
         reverse("full-text-html", kwargs={"document_uri": document.uri}),
     )
 
-    preferred_ncn: Identifier | None = document.identifiers.preferred(type=NeutralCitationNumber)
+    description_string = f"""EUI: {editor_html_url}
```
dragon-dxw, line 93: I'd be tempted to write:

identifiers = "\n\t".join([f"\n\t{identifier.schema.name}: {identifier.value}" for identifier in document.identifiers.values()])

and then insert that into the f string; possibly also using variables and f strings instead of format throughout.

judgments/utils/link_generators.py
```diff
@@ -80,31 +79,35 @@ def build_raise_issue_email_link(
...
+
+Identifiers:"""
```
dragon-dxw, line 95: Could do with a test.

dragon-dxw (approved):

PR #1867, opened 2025-01-23
[FCL-626] Reinstate updating a document URI on NCN update

jacksonj04:
We haven't yet implemented the ingester change to provide UUID-based document URIs, but editors are still receiving documents at `/failure` URIs. Since the new code won't change these (by design), we need to temporarily reinstate the functions which update the URI when the NCN changes.

As part of this we refactor out the logic in `EditJudgmentView.post` which processes the actual change into a new `update_ncn_of_document` method.

### DIFF
--- a/judgments/utils/aws.py
+++ b/judgments/utils/aws.py
@@ -1,73 +1,9 @@
 import time
 
 import boto3
-import botocore.client
 import environ
 
 env = environ.Env()
-
-
-def copy_assets(old_uri, new_uri):
-    client = create_s3_client()
-    bucket = env("PRIVATE_ASSET_BUCKET")
-    old_uri = uri_for_s3(old_uri)
-    new_uri = uri_for_s3(new_uri)
-
-    response = client.list_objects(Bucket=bucket, Prefix=old_uri)
-
-    for result in response.get("Contents", []):
-        old_key = str(result["Key"])
-        new_key = build_new_key(old_key, new_uri)
-        if new_key is not None:
-            source = {"Bucket": bucket, "Key": old_key}
-            client.copy(source, bucket, new_key)
-
-
-def build_new_key(old_key, new_uri):
-    old_filename = old_key.rsplit("/", 1)[-1]
-
-    if old_filename.endswith((".docx", ".pdf")):
-        new_filename = new_uri.replace("/", "_")
-        return f"{new_uri}/{new_filename}.{old_filename.split('.')[-1]}"
-    else:
-        return f"{new_uri}/{old_filename}"
-
-
-def create_aws_client(service: str):  # service
-    """@param: service The AWS service, e.g. 's3'"""
-    aws = boto3.session.Session(
-        aws_access_key_id=env("AWS_ACCESS_KEY_ID", default=None),
-        aws_secret_access_key=env("AWS_SECRET_KEY", default=None),
-    )
-    return aws.client(
-        service,
-        endpoint_url=env("AWS_ENDPOINT_URL", default=None),
-        region_name=env("PRIVATE_ASSET_BUCKET_REGION", default=None),
-        config=botocore.client.Config(signature_version="s3v4"),
-    )
-
-
-def create_s3_client():
-    return create_aws_client("s3")
-
-
-def uri_for_s3(uri: str):
-    return uri.lstrip("/")
-
-
-def generate_signed_asset_url(key: str):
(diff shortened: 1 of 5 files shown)


judgments/views/judgment_edit.py
```diff
@@ -30,6 +33,38 @@ def verify_stub_not_used(old_uri, new_citation) -> None:
         raise StubAlreadyUsedError(msg)
 
 
+def update_ncn_of_document(document: Document, new_neutral_citation_number_string: str) -> None:
+    """Given a documen and a new NCN string, validate that NCN and perform necessary operations to update the document with it."""
```
dragon-dxw, line 37: trivial typo: `documen`

judgments/views/judgment_edit.py
```diff
@@ -30,6 +33,38 @@ def verify_stub_not_used(old_uri, new_citation) -> None:
...
+    # Perform sense-check on document type
+    if document.document_noun != "judgment":
+        msg = f"Cannot update the NCN of non-judgment at {document.uri}"
+        raise CannotUpdateNCNOfNonJudgment(msg)
```
dragon-dxw, line 41: This new exception needs to get trapped further upstream. We probably need a "ExpectedUIError" class for "this is an editor problem"

judgments/views/judgment_edit.py
```diff
@@ -30,6 +33,38 @@ def verify_stub_not_used(old_uri, new_citation) -> None:
...
+    new_uri_raw = neutral_url(new_neutral_citation)
+    if new_uri_raw is None:
+        msg = f"Unable to form new URI from neutral citation: {new_neutral_citation}"
+        raise NeutralCitationToUriError(msg)
+
+    # Confirm that the identifier isn't used elsewhere
+    verify_stub_not_used(document.uri, new_neutral_citation)
+
+    # Set neutral citation in the identifiers
+    document.identifiers.delete_type(NeutralCitationNumber)
```
dragon-dxw, line 56: [not blocking this PR] Do we want a function that combines these?

judgments/utils/aws.py
```diff
@@ -7,6 +8,37 @@
...
+
+    for result in response.get("Contents", []):
+        old_key = str(result["Key"])
+        new_key = build_new_key(old_key, new_uri)
+        if new_key is not None:
+            try:
+                source = {"Bucket": bucket, "Key": old_key}
+                client.copy(source, bucket, new_key)
+            except botocore.client.ClientError as e:
+                logging.warning(
```
dragon-dxw, line 27: Will this trigger rollbar? If not, we should think of a way of surfacing this.

judgments/utils/aws.py
```diff
@@ -7,6 +8,37 @@
...
+                    f"Unable to copy file {old_key} to new location {new_key}, error: {e}",
+                )
+
+
+def build_new_key(old_key, new_uri):
```
dragon-dxw, line 32: This might benefit from the use of some of those types we just added to APIClient

judgments/views/judgment_edit.py
```diff
@@ -114,7 +137,8 @@ def post(self, request, *args, **kwargs):
         except MarklogicAPIError as e:
             messages.error(request, f"There was an error saving the Document: {e}")
 
-        invalidate_caches(judgment.uri)
+        if not settings.DEBUG:
```
dragon-dxw, line 140: [not blocking this PR] This feels like a thing the API client should know and care about, rather than the editor UI.

dragon-dxw (approved):

dragon-dxw:
Code Climate seems to be blocking merge. I don't seem to be able to meaningfully interact with CodeClimate to see what's wrong.

judgments/views/judgment_edit.py
jacksonj04: I've made all these a new `NCNValidationException` which we catch. We no longer catch exceptions bubbled from lower-level API methods, because we shouldn't be calling them in the first place.

judgments/views/judgment_edit.py
jacksonj04: I don't think so; this is the only real place where we'll ever have to nuke and totally replace an ID, and it's temporary.

judgments/utils/aws.py
jacksonj04: Agree this shouldn't be a try/catch at all; if we try this and it fails it's a real exception and we want to know about it.

PR #1868, opened 2025-01-23
Explicit 404 for favicon.ico rather than 500ing due to invalid NCN URL

dragon-dxw:
A large number of InvalidDocumentURIExceptions are browsers requesting favicon.ico. Explicitly return a 404 rather than generating an exception.

https://app.rollbar.com/a/dxw/fix/item/tna-caselaw-editor-ui/195/occurrence/425839049443#occurrences


jacksonj04 (approved):
The other option would be providing a favicon!

PR #1922, opened 2025-02-20
Remove unsafe legacy unused hold endpoint

anthonyhashemi:
## Changes in this PR:

Remove unsafe legacy unused hold endpoint.
Screenshot 2025-02-20 at 15 52 00

Prompted by seeing the security issue; https://github.com/nationalarchives/ds-caselaw-editor-ui/security/code-scanning/31 and then realising that we have 2 "hold" deifnitions in our url patterns. The newer one is fine and is what we use as it is defined first in the patterns list and django picks the first pattern that matches (https://docs.djangoproject.com/en/5.1/ref/urls/#:~:text=When%20processing%20a%20request%2C%20Django,parameters%2C%20or%20the%20domain%20name.)

https://national-archives.atlassian.net/browse/FCL-698


Donna-H (approved):
okay for removal for unused hold endpoint.

PR #2005, opened 2025-04-22
Report MoveJudgmentError to users

dragon-dxw:
## Changes in this PR:
Whilst we should have a concept in the Editor of a ReportableError (for anticipated user-data problems) let's just add it to the list of things to display to editors. 

https://app.rollbar.com/a/dxw/fix/item/tna-caselaw-editor-ui/228/occurrence/432777794764#detail

### After


jacksonj04:
Closing at this is very edgy, most cases will already be caught, and we're already planning to remove this entire behaviour.

jacksonj04:
Closing at this is very edgy, most cases will already be caught, and we're already planning to remove this entire behaviour.

PR #2007, opened 2025-04-23
[FCL-808] Remove the EUI trying to move documents on NCN change

jacksonj04:
This is no longer the desired behaviour; all document URIs are frozen since URL resolution is now handled via the identifiers layer.

For existing documents, this should already be the right NCN (if not we'll resolve this in a future exception report). For new documents, this should be a non-semantic UUID.

This change does remove the protections against multiple documents having the same NCN, since this is no longer a problem at a database level.

## Jira

FCL-808

### DIFF
--- a/judgments/views/judgment_edit.py
+++ b/judgments/views/judgment_edit.py
@@ -1,76 +1,41 @@
 import datetime
 
 from caselawclient.Client import MarklogicAPIError
-from caselawclient.models.documents import Document, DocumentURIString
-from caselawclient.models.identifiers.neutral_citation import NeutralCitationNumber, NeutralCitationNumberSchema
-from caselawclient.types import DocumentIdentifierSlug
+from caselawclient.models.documents import Document
+from caselawclient.models.identifiers.neutral_citation import NCNValidationException, NeutralCitationNumber
 from django.conf import settings
 from django.contrib import messages
 from django.http import HttpResponseRedirect
 from django.urls import reverse
 from django.views.generic import View
-from ds_caselaw_utils import neutral_url
-from ds_caselaw_utils.types import NeutralCitationString
 
-from judgments.utils import api_client, update_document_uri
+from judgments.utils import api_client
 from judgments.utils.aws import invalidate_caches
 from judgments.utils.view_helpers import get_document_by_uri_or_404
 
 
-class NCNValidationException(Exception):
+class CannotUpdateNCNOfNonJudgment(Exception):
     pass
-
-
-class SlugAlreadyUsedError(NCNValidationException):
-    pass
-
-
-class CannotUpdateNCNOfNonJudgment(NCNValidationException):
-    pass
-
-
-class NeutralCitationToUriError(NCNValidationException):
-    pass
-
-
-def verify_slug_not_used(identifier_slug: DocumentIdentifierSlug) -> None:
-    if resolutions := api_client.resolve_from_identifier_slug(identifier_slug):
-        msg = f"At least one identifier for slug {identifier_slug} already exists, {resolutions}"
-        raise SlugAlreadyUsedError(msg)
 
 
 def update_ncn_of_document(document: Document, new_neutral_citation_number_string: str) -> None:
     """Given a document and a new NCN string, validate that NCN and perform necessary operations to update the document with it."""
+
     # Perform sense-check on document type
     if document.document_noun != "judgment":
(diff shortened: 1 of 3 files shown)


judgments/views/judgment_edit.py
```diff
@@ -41,37 +40,18 @@ def verify_slug_not_used(identifier_slug: DocumentIdentifierSlug) -> None:
 
 def update_ncn_of_document(document: Document, new_neutral_citation_number_string: str) -> None:
     """Given a document and a new NCN string, validate that NCN and perform necessary operations to update the document with it."""
-    # Perform sense-check on document type
-    if document.document_noun != "judgment":
```
anthonyhashemi, line 45: why is this check not necessary anymore?

judgments/views/judgment_edit.py
```diff
@@ -41,37 +40,18 @@ def verify_slug_not_used(identifier_slug: DocumentIdentifierSlug) -> None:
...
-
-    # Confirm that the identifier isn't used elsewhere
-    new_ncn = NeutralCitationNumber(value=new_neutral_citation)
-    # TODO: change types in API client to return DocumentIdentifierSlug and remove this coercion
-    verify_slug_not_used(DocumentIdentifierSlug(NeutralCitationNumberSchema.compile_identifier_url_slug(new_ncn.value)))
-
     # Set neutral citation in the identifiers
     document.identifiers.delete_type(NeutralCitationNumber)
-    document.identifiers.add(new_ncn)
+    document.identifiers.add(NeutralCitationNumber(value=new_neutral_citation))
```
anthonyhashemi, line 49: can use `new_neutral_citation` from line 46

judgments/views/judgment_edit.py
```diff
@@ -41,37 +40,18 @@ def verify_slug_not_used(identifier_slug: DocumentIdentifierSlug) -> None:
...
     document.save_identifiers()
 
     # Set neutral citation in the document XML
     api_client.set_judgment_citation(document.uri, new_neutral_citation)
```
anthonyhashemi, line 53: should the above logic be encapsulated by `api_client.set_judgment_citation(document.uri, new_neutral_citation)` I wonder / be encapsulated by soome method in the apiclient, or is this eui specific?

judgments/views/judgment_edit.py
dragon-dxw: This was a guard against having to handle press summaries when we didn't have code for correctly renaming them.

judgments/views/judgment_edit.py
dragon-dxw, line 49: ... but this is now wrong because there's no guarantee that this document is in fact a judgment (implied by having an NeutralCitationNumber rather than a press summary version of this construct)

We should probably have hardcoded checks that we don't set NCNs on non-judgments.

judgments/views/judgment_edit.py
dragon-dxw: There are a bunch of subtleties that will trip us up if we try to make this work with press summaries without careful thought -- recommend leaving this in for now.

Things that won't work out of the box
* modifying the XML
* modifying the identifiers

dragon-dxw (changes requested):
This isn't ready for making it work with press summaries.

judgments/views/judgment_edit.py
jacksonj04: This has been reinstated, we should make sure there are stricter identifier integrity rules in place in APIC first, and this isn't a feature regression.

judgments/views/judgment_edit.py
jacksonj04: It's actually the other way round; it _should_ be the case that modifying any of the identifiers will in turn modify the XML. However, this has been left in a mildly (albeit not severely) inconsistent state thanks to `set_judgment_citation` not reliably being called in all contexts.

The plan at the moment is we're going to build app-side functionality for cleaning up a document's metadata in the XML when `save_identifiers` is called, then running a maintenance task to make sure all XML is current.

judgments/views/judgment_edit.py
jacksonj04: This is now guarded against again.

anthonyhashemi (approved):
The PR looks much simpler and focussed now; does what it says on the tin. Nice ✅

dragon-dxw (approved):

PR #2036, opened 2025-05-02
FCL-822 Use doc.content_as_html not doc.body's

dragon-dxw:
Uses API Client 37 to correctly write the image names.
https://national-archives.atlassian.net/browse/FCL-822

Note that this also changes prefixes to not include the trailing slash: `/signed-assets` is now correct, whereas it was `/signed-assets/`; the enviroment variable will need changing.

### DIFF
--- a/judgments/utils/view_helpers.py
+++ b/judgments/utils/view_helpers.py
@@ -91,15 +91,11 @@ class DocumentView(TemplateView):
 
         version_uri = self.request.GET.get("version_uri", None)
 
-        xslt_image_location = env("XSLT_IMAGE_LOCATION", default=None)
-
         if version_uri:
             context["current_version_number"] = extract_version_number_from_filename(version_uri)
-            context["document_html"] = get_document_by_uri_or_404(version_uri).body.content_as_html(
-                image_base_url=xslt_image_location,
-            )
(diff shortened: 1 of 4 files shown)


jacksonj04 (approved):

judgments/utils/view_helpers.py
```diff
@@ -91,15 +91,11 @@ def get_context_data(self, **kwargs):
...
         version_uri = self.request.GET.get("version_uri", None)
 
-        xslt_image_location = env("XSLT_IMAGE_LOCATION", default=None)
-
         if version_uri:
             context["current_version_number"] = extract_version_number_from_filename(version_uri)
-            context["document_html"] = get_document_by_uri_or_404(version_uri).body.content_as_html(
-                image_base_url=xslt_image_location,
-            )
+            context["document_html"] = get_document_by_uri_or_404(version_uri).content_as_html()
```
anthonyhashemi, line 96: this is the line of code which is apparently untested according to the codeclimate chrome extension I have installed. 

Adding a test for this logic should get the check passing

dragon-dxw: thank you! -- I don't know how to get the codeclimate extension to work properly, and I've spent too long today trying to work out what the problem is.

dragon-dxw: image

PR #2058, opened 2025-05-15
[FCL-49] Add management command to assign missing FCLIDs

jacksonj04:
This is a maintenance command intended to be run manually. It will loop over the query for documents without an FCLID until there are none left, and then for each document in the query will attempt to mint a new FCLID.

### DIFF
--- a/judgments/management/commands/assign_missing_fclids.py
+++ b/judgments/management/commands/assign_missing_fclids.py
@@ -1,0 +1,45 @@
+from caselawclient.models.identifiers.exceptions import IdentifierValidationException
+from caselawclient.types import MarkLogicDocumentURIString
+from django.core.management.base import BaseCommand
+
+from judgments.utils import api_client
+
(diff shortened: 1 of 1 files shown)


judgments/management/commands/assign_missing_fclids.py
```diff
@@ -0,0 +1,45 @@
...
+        while total > 0:
+            loop += 1
+
+            documents_missing_fclid = api_client.get_missing_fclid(maximum_records=NUMBER_PER_LOOP)
+            total = len(documents_missing_fclid)
+
+            for i, document in enumerate(documents_missing_fclid):
+                doc_count += 1
+
+                prefix_string = f"L{loop}, {i + 1}/{total} ({doc_count})"
```
dragon-dxw, line 27: I don't get the difference between `i+1` and `doc_count`

dragon-dxw (approved):

PR #2061, opened 2025-05-15
Add multifactor authentication

dragon-dxw:
## Changes in this PR:
Add multifactor authentication

This won't suddenly spring MFA on anyone; they'll have to visit /accounts/mfa to set it up.

Does require a database migration.

https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136/backlog?selectedIssue=FCL-131

## Screenshots of UI changes:
image

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -74,6 +74,7 @@ THIRD_PARTY_APPS = [
     "allauth",
     "allauth.account",
     "allauth.socialaccount",
+    "allauth.mfa",
     "stronghold",
     "waffle",
     "widget_tweaks",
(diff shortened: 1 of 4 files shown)


config/settings/base.py
```diff
@@ -282,17 +283,17 @@
 # https://django-allauth.readthedocs.io/en/latest/configuration.html
 ACCOUNT_ADAPTER = "judgments.account_adapter.NoNewUsersAccountAdapter"
 ACCOUNT_ALLOW_REGISTRATION = False
-ACCOUNT_AUTHENTICATION_METHOD = "username"
-ACCOUNT_EMAIL_REQUIRED = True
+ACCOUNT_LOGIN_METHODS = {"username"}
```
dragon-dxw, line 286: These two changes were recommended by the software itself.

config/settings/base.py
```diff
@@ -282,17 +283,17 @@
...
+MFA_ADAPTER = "allauth.mfa.adapter.DefaultMFAAdapter"
 
 
 # Your stuff...
 # ------------------------------------------------------------------------------
 
 STRONGHOLD_PUBLIC_URLS = (
-    r"^/accounts/login",
-    r"^/accounts/password/reset/",
+    r"^/accounts/",
```
dragon-dxw, line 296: This has security implications, but I'm pretty sure they're fine.

ds_caselaw_editor_ui/templates/allauth/layouts/base.html
```diff
@@ -0,0 +1,87 @@
+{% extends "layouts/base.html" %}
```
dragon-dxw, line 1: This is copy-pasted from https://github.com/pennersr/django-allauth/blob/main/allauth/templates/allauth/layouts/base.html with the exception of the first line and the html lang tag as recommended by linter.

dragon-dxw:
We could probably turn on passkeys if we wanted to.
https://docs.allauth.org/en/dev/mfa/webauthn.html

config/settings/base.py
jacksonj04: What implications?

PR #2073, opened 2025-05-21
Fcl 847/use frontend tna colours

jlhdxw:
## Changes in this PR:

Updates colours from the new ones in FCL frontend and removes a bunch of unused stuff. Also clean up some styles that weren't quite right.

https://national-archives.atlassian.net/browse/FCL-847


Terry-Price (approved):
after a conversation with @Floppy - happy to include unused colours due to technical constraints.

PR #2099, opened 2025-06-03
[FCL-881] editors are not able to view document history pages

jacksonj04:
Bump to fixed version of API Client to allow history to be loaded for pages with UUID-based URIs. As a bonus action, promote structured history view out of labs.

### DIFF
--- a/judgments/tests/test_document_history.py
+++ b/judgments/tests/test_document_history.py
@@ -4,57 +4,17 @@ from unittest.mock import patch
 
 from caselawclient.client_helpers import VersionAnnotation, VersionType
 from caselawclient.factories import DocumentBodyFactory, JudgmentFactory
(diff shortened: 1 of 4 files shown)


judgments/tests/test_document_history.py
```diff
@@ -4,57 +4,17 @@
...
 from django.test import TestCase
 from django.urls import reverse
 from factories import DocumentVersionFactory
-from waffle.testutils import override_flag
 
 from judgments.views import document_history
 from judgments.views.document_history import DocumentHistorySequencer
 
 
-class TestDocumentHistory(TestCase):
```
anthonyhashemi, line 19: Took me a second but for anyone else these aren't needed as they were for the old history page before the toggle

anthonyhashemi (approved):
Nice!

PR #2217, opened 2025-08-04
You must be logged in to visit urls with `/check` prefix unless they're literally `/check`

dragon-dxw:
We noticed an attacker got a 500 trying to visit `/checkout/paypal.env`, rather than a 403 or redirect to login.
We were allowing any urls starting /check to be visited when not logged in, rather than just the /check endpoint.

### DIFF
--- a/judgments/tests/tests.py
+++ b/judgments/tests/tests.py
@@ -28,3 +28,17 @@ class TestSearchResults(TestCase):
                 page=1,
             ),
         )
+
+
+class TestCheckPrefixUrls(TestCase):
+    def test_just_check_ok(self):
+        """The /check endpoint can be used when not logged in"""
+        response = self.client.get("/check")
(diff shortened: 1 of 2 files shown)


jacksonj04 (approved):

judgments/tests/tests.py
```diff
@@ -28,3 +28,17 @@ def test_oldest(self, mock_search):
...
+
+
+class TestCheckPrefixUrls(TestCase):
+    def test_just_check_ok(self):
+        """The /check endpoint can be used when not logged in"""
+        response = self.client.get("/check")
+        assert response.status_code == 200
+        assert b'"OK"' in response.content
+
+    def test_check_prefix_not_ok(self):
```
anthonyhashemi, line 40: amazing, hanks

PR #2219, opened 2025-08-06
Do not send unenrichable documents to enrichment

dragon-dxw:
## Changes in this PR:

Calling `document.enrich(allow_failures=True)` will return `None` rather than raise an exception if it's not reasonable to enrich the document. This is good for both the cronjob (which will silently ignore and skip) and the UI (which will produce a sensible error)

It's possible that we should instead mark the document so that it is never reconsidered for enrichment in some way.

https://national-archives.atlassian.net/browse/FCL-1087

### DIFF
--- a/judgments/management/commands/enrich_next_in_reenrichment_queue.py
+++ b/judgments/management/commands/enrich_next_in_reenrichment_queue.py
@@ -35,5 +35,5 @@ class Command(BaseCommand):
             document = api_client.get_document_by_uri(document_uri)
 
             self.stdout.write(f"Sending document {document.name} to enrichment...")
-            document.enrich()
+            document.enrich(accept_failures=True)
             time.sleep(3)
--- a/judgments/views/enrich.py
+++ b/judgments/views/enrich.py
@@ -8,7 +8,7 @@ from judgments.utils.view_helpers import get_document_by_uri_or_404
 def enrich(request):
     document_uri = request.POST.get("document_uri", None)
     document = get_document_by_uri_or_404(document_uri)
-    enrichment_triggered = document.enrich(accept_failures=True)
+    enrichment_triggered = document.enrich(accept_failures=True, even_if_recent=True)
     if not enrichment_triggered:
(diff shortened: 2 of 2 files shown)


anthonyhashemi:
Also agree that we could mark documents that should not be enriched, but let's think of a good way of doing that.

For now, we can definitely just ignore documents without a source docx, which I believe we do for reparse

judgments/management/commands/enrich_next_in_reenrichment_queue.py
```diff
@@ -35,5 +35,5 @@ def handle(self, *args, **options):
             document = api_client.get_document_by_uri(document_uri)
 
             self.stdout.write(f"Sending document {document.name} to enrichment...")
-            document.enrich()
+            document.enrich(accept_failures=True)
```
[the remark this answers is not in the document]
dragon-dxw: Essentially, yes; we could do that instead.

anthonyhashemi:
> Not necessarily documents without a docx; some early documents were uploaded without docx but with full XML, so can be enriched but not reparsed.

Oh I see, fair point!

anthonyhashemi (approved):

judgments/views/enrich.py
```diff
@@ -8,7 +8,7 @@
 def enrich(request):
     document_uri = request.POST.get("document_uri", None)
     document = get_document_by_uri_or_404(document_uri)
-    enrichment_triggered = document.enrich()
+    enrichment_triggered = document.enrich(accept_failures=True)
```
anthonyhashemi, line 11: I guess we already have a test for when document.enrich returns falsey

judgments/management/commands/enrich_next_in_reenrichment_queue.py
anthonyhashemi, line 38: could add a test to demonstrate that we want it to continue to try enrich the next one even if one doc errors on doc.enrich

PR #2799, opened 2026-04-13
ci: add production smoke test and modernise docker-compose

anthonyhashemi:
## Changes in this PR:

Get EUI inline with changes to dockerfile/compose made in PUI for consistency.

- Replace build-only CI step with compose-based production smoke test
- Add docker-compose.prod-test.yml overlay targeting production Dockerfile
- Add MarkLogic and postgres healthchecks with service_healthy depends_on
- Replace hardcoded env vars and env_file with ${VAR:-default} syntax
- Fix Dockerfile permissions: app owned by root, only runtime dirs chowned
- Add node_modules to .dockerignore, --unsafe to check-yaml pre-commit
- Update README with compose files, MarkLogic, and network docs

https://national-archives.atlassian.net/browse/FCLC-624

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,4 +7,5 @@
 .pre-commit-config.yaml
 .readthedocs.yml
 .travis.yml
+node_modules
 venv
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
           name: django-image
           path: /tmp/django-image.tar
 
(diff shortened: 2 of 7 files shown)


jacksonj04 (approved):

jacksonj04 (changes requested):
Looks like this needs the e2e configuration tweaking as well, but otherwise this is a great chunk of cleaning up cruft!

jacksonj04 (approved):

PR #2800, opened 2026-04-13
Fclc 1772/add vcr for e2e

jlhdxw:
## Changes in this PR:

Adds VCR and backfills lots of e2e tests now we aren't loading dynamic data when running these.

This will allow us to slowly migrate to jinja with more confidence.

https://national-archives.atlassian.net/browse/FCLC-1772


.github/workflows/ci.yml
```diff
@@ -276,37 +273,16 @@ jobs:
...
 
-      - name: Run marklogic docker container
-        run: docker compose up -d --wait
-        working-directory: "./marklogic"
-
-      - name: Deploy marklogic configuration
-        run: gradle mlDeploy
-        working-directory: "./marklogic"
-
-      - name: Populate marklogic DB with test data
```
anthonyhashemi, line 303: So was this not adding data to the marklogic instance at all before? Or was because the data is coming from staging or whatever the env vars point to and so isnt necessarily reproducible?

anthonyhashemi (approved):
Thanks for finding and implementing this solution, it is a really smart approach to achieve confidence in our UI without needing a whole large marklogic instance.

I know we already have implemented this in PUI and this is just an equivalent change to EUI and I agree this solves a real problem, but I just wanted to flag that it does feel like a lot of code.

The amount of infra/env‑gated code needed to support VCR suggests the snapshots are living at too high a level.

Longer term, I feel like we could look to move most visual regression coverage to storybook to do component level tests, where test data is simple to provide.

That way we'd reduce this complexity and leave just a small number of E2E tests for genuinely end‑to‑end flows with some basic data seeded in a Marklogic container.

But yeah, I am very happy with this as a solution for now as it gives us great confidence as we start to migrate to jinja components which will in turn leads us to being able to do storybook visual component testing.

.github/workflows/ci.yml
jacksonj04: [This script](https://github.com/nationalarchives/ds-caselaw-marklogic/blob/main/development_scripts/populate_from_caselaw.py) comes from the MarkLogic repo, and just loads up some fixture files. It should be reproducible, _but_ doesn't have good/complex enough fixtures to exercise things we necessarily care about.

anthonyhashemi: ah right, thanks! As mentioned in my other comment, we probably would still like some e2e test connecting to marklogic for verification that we have valid connection but it should hopefully be a very minimal set of tests.  I will wait for this to be merged in so that my PR stops failing: https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/2799 It was failing because it adds marklogic as a container to the docker-compose file like I did in PUI to make it possible for devs to spin up an env with no extra steps - just that one compose file and no access to staging or another repo. The data there is currently empty, but a followup ticket may involve incorporating a version of this populate script for local dev.

PR #2868, opened 2026-04-24
Fclc 1825/migrate views to jinja

jlhdxw:
## Changes in this PR:

Migrates the following pages to jinja:

 - Publish
 - Publish success
 - Unpublish
 - Unpublish success
 - Downloads
 - History
 - Hold
 - Hold success
 - Unhold
 - Unhold success
 - Delete
 - Upload
 - Identifiers
 -  Identifiers add
 - Identifiers delete
 - Reports 
 - Reports -> Awaiting parse (this breaks on staging)
 - Reports -> Bulk reparse run log list
 - Reports -> Bulk reparse run log detail
 - Reports -> Awaiting enrichment  (this breaks on staging)
 - Reports -> Locked documents - 

Still to do:

 - Upload success ( in another PR, though will likely need to be updated to use the new components)
 - Labs (do we still use this?)
 - Associated documents
 - Stats (is this still used?)

 - Check

https://national-archives.atlassian.net/browse/FCLC-1825


judgments/views/reports.py
```diff
@@ -71,13 +77,21 @@ class BulkReparseRunLogDetailView(DetailView):
...
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        target_enrichment_version = api_client.get_highest_enrichment_version()
-        target_parser_version = api_client.get_highest_parser_version()
+        try:
+            target_enrichment_version = api_client.get_highest_enrichment_version()
+        except (TypeError, ValueError):
+            target_enrichment_version = (0, 0)
```
anthonyhashemi, line 89: does this fallback make sense?

judgments/views/document_identifiers.py
```diff
@@ -18,14 +18,25 @@
...
+
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
         context["view"] = "document_identifiers"
+        preferred_identifier = self.document.identifiers.preferred()
 
-        return context
+        if preferred_identifier is not None:
+            context["preferred_identifier_name"] = preferred_identifier.schema.name
```
anthonyhashemi, line 31: I guess all of this logic here vs means the jinja template is simpler than the old django equivalent

judgments/views/document_full_text.py
```diff
@@ -24,18 +23,6 @@ def dispatch(self, request, *args, **kwargs):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        document = context["document"]
-
-        if document.has_ever_been_published:
```
anthonyhashemi, line 29: Does this mean the jinja template is now more complicated than the old django eauivalent (which had been deleted in a previous PR)? If so, why?

judgments/utils/view_helpers.py
```diff
@@ -151,6 +152,16 @@ def get_context_data(self, **kwargs):
 
         context["preferred_ncn"] = self.document.identifiers.preferred(type=NeutralCitationNumber)
 
+        if self.document.has_ever_been_published:
```
anthonyhashemi, line 155: oh I see, you moved it to this mixin! makes sense

judgments/tests/test_document_identifiers.py
```diff
@@ -144,7 +144,7 @@ def test_identifier_view_delete_disablement_published(self, document_type, docum
             <td><code>[2025] UKSC 123</code></td>
             <td><code>/uksc/2025/123</code></td>
             <td>False</td>
-            <td><a class="button-danger button-small" disabled aria-disabled="true">Delete</a></td>
+            <td><button class="button button--danger button--small" disabled="true">Delete</button></td>
```
anthonyhashemi, line 147: makes sense to migrate to use button where it is a button. Do we not have a more specific custom button component or the raw one is sufficient?

judgments/templatetags/document_utils.py
```diff
@@ -32,6 +32,8 @@ def render_json(json_string):
 @register.filter
 def get_dict_key_with_hyphen(the_dict, key):
     """Django templates don't support using hyphens in variable names (eg `the_dict.key-with-hyphen`), so this provides a workaround."""
+    if not the_dict:
```
anthonyhashemi, line 35: Just because it is a large pr touching many pages with mainly non-concerning migrations, I'm just commenting on the logic changes that I have any questions on.

Do we have a test for this and what was this breaking before?

e2e_tests/test_judgment_page.py
```diff
@@ -15,6 +15,11 @@ def test_judgment_identifiers(authenticated_page: Page):
     assert_matches_snapshot(authenticated_page, "judgment_identifiers_page")
 
 
+def test_judgment_identifiers_add(authenticated_page: Page):
+    authenticated_page.goto(f"{JUDGMENT_URI}/identifiers/add")
```
anthonyhashemi, line 19: thanks for adding

although I wonder if there is an automatic way for us to cover all the pages without having to remember if there is a new page? 🤔

ds_caselaw_editor_ui/templates/includes/judgment/history/structured_submission.jinja
```diff
@@ -0,0 +1,36 @@
...
+  <div id="submission-{{ submission.submission_sequence_number }}"
+       {% if not expand %}hidden{% endif %}>
+    <div class="table">
+      <table class="document-history__event-table">
+        <tr>
+          <th>Change made</th>
+          <th>Changed at</th>
+          <th>Changed by</th>
+          {% for event in submission.events | reversed %}
+            {{ event_line(event=event, document_uri=document_uri, user=user) }}
```
anthonyhashemi, line 30: amazing use of componentization 😄 this page is so much more organised now

ds_caselaw_editor_ui/templates/judgment/publish-success.jinja
```diff
@@ -0,0 +1,83 @@
...
+         target="_blank"
+         rel="noopener noreferrer">View the published document <span class="icon icon--sm icon--link icon__arrow-up-right-square"></span></a>
+    </p>
+    <p>
+      <a href="{{ email_confirmation_link }}">Email the submitter</a>
+    </p>
+  </div>
+{% endblock content %}
+{% block footer_javascript %}
+  <script src="https://cdn.jsdelivr.net/npm/tsparticles-preset-confetti@2/tsparticles.preset.confetti.bundle.min.js"></script>
```
anthonyhashemi, line 20: Even though our code does not process user input, or affect application state or touch auth / sensitive data, this feels like a potential supply chain attack risk if the cdn gets compromised.

At minimum we should:
1. Pin an exact version (e.g. https://cdn.jsdelivr.net/npm/tsparticles-preset-confetti@2.11.1/dist/tsparticles.preset.confetti.bundle.min.j )
2. Add Subresource Integrity `integrity="<hash>", crossorigin="anonymous"`
3. If we have not already restrict what domains we can communicate with via a CSP

Or we could go further and install with npm

ds_caselaw_editor_ui/templates/judgment/identifiers.jinja
```diff
@@ -0,0 +1,101 @@
+{% extends "layouts/judgment.jinja" %}
```
anthonyhashemi, line 1: this is now the largest template at 101 lines, is there anything else that makes sense to break out into a component here? Might not be, but just flagging

anthonyhashemi (approved):
Great work, approving assuming all comments addressed and @Terry-Price gives an OK on all the very minor style tweaks visible in the visual regression snapshots.

judgments/views/reports.py
jlhdxw: I'm not sure - I could set to something else like "None" and handle that differently?

It looks like this:

image

This page fails to load on staging

judgments/views/document_identifiers.py
jlhdxw: That and django falls back and hides when you try to access data that doesn't exist.

judgments/views/document_full_text.py
jlhdxw: As you've seen, it's moved to the helper as we had this logic duplicated so I figured I'd make the views simpler.

judgments/tests/test_document_identifiers.py
jlhdxw: This is in the test - we're using the jinja component in the view (which is why this renders as a button now rather than an an `a` tag without a href (which isn't a valid element)

judgments/templatetags/document_utils.py
jlhdxw: Django was hiding the issue previously, but jinja raises an error for this. There's no existing test for this but I'll add one

ds_caselaw_editor_ui/templates/judgment/publish-success.jinja
jlhdxw: I'll add a ticket to do this as a follow up

anthonyhashemi: cheers

ds_caselaw_editor_ui/templates/judgment/identifiers.jinja
jlhdxw: It's all only used here in this template, so for now probably not worth separating anything out. It's about the same as the html version

PR #2889, opened 2026-04-30
FCLC-1864 | remove labs

jlhdxw:
## Changes in this PR:

Remove the labs page as it's unused.

https://national-archives.atlassian.net/browse/FCLC-1864


dragon-dxw (approved):
If we're agreed we're okay with deleting labs, the PR is good.

PR #2898, opened 2026-05-01
FCLC-1866 | remove stats page

jlhdxw:
## Changes in this PR:

Remove unused stats page

https://national-archives.atlassian.net/browse/FCLC-1866


judgments/urls.py
```diff
@@ -101,13 +100,6 @@
         reports.LockedDocuments.as_view(),
         name="report_locked_documents",
     ),
-    # Stats
-    path("stats", Stats.as_view(), name="stats"),
-    path(
-        "stats/combined-csv",
-        stream_combined_stats_table_as_csv,
```
dragon-dxw, line 108: we probably want to get rid of this function too.

jlhdxw: I removed the whole view file, so this function is deleted

dragon-dxw (approved):

PR #2910, opened 2026-05-06
FCLC-1872 | migrate bulk reparse run log detail to jinja

jlhdxw:
## Changes in this PR:

Migrate bulk reparse run log detail to jinja

https://national-archives.atlassian.net/browse/FCLC-1872


dragon-dxw (approved):
Surprised to not see a HTML file deletion.

PR #2950, opened 2026-05-18
migrate to lxml

dragon-dxw:
FCLC-1930

Support the breaking changes in the API client's most recent version which migrates from xml/defusedxml to lxml

### After

### DIFF
--- a/judgments/tests/test_stub.py
+++ b/judgments/tests/test_stub.py
@@ -3,10 +3,10 @@ from unittest.mock import ANY, MagicMock, call, patch
 
 from caselawclient.Client import ROOT_DIR
 from caselawclient.models.judgments import Judgment
-from defusedxml import ElementTree
 from django.contrib.auth.models import User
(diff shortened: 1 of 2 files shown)


judgments/tests/test_stub.py
```diff
@@ -73,7 +73,7 @@ def test_judgment_stub_view(self):
     @patch("judgments.views.stub.api_client.set_property")
     def test_judgment_stub_post(self, mock_set_property, mock_insert_xml, mock_render_stub, mock_uuid, mock_courts):
         judgment_template_path = Path(ROOT_DIR) / "models" / "documents" / "templates" / "judgment.xml"
-        with (judgment_template_path).open("r") as f:
+        with (judgment_template_path).open("rb") as f:
```
dragon-dxw, line 76: lxml is picky about not allowing unicode strings with unicode declarations, so we make everything bytes.

dynamictulip (approved):
LGTM

PR #2951, opened 2026-05-19
NCN on stub page

dragon-dxw:
This needs to change in sync with the version of the API client as the spec for the data passed is in the API client.

This requires https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/1782 merged in the API client.

## Changes in this PR:
image

image


anthonyhashemi (approved):
Asumming tests pass when the corresponding apiclient update is released this looks good https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/1782#pullrequestreview-4580069570

PR #3069, opened 2026-06-18
Fclc 2032/update metadata styling

jlhdxw:
## Changes in this PR:

Update metadata styling.

https://national-archives.atlassian.net/browse/FCLC-2032


ds_caselaw_editor_ui/templates/components/document_metadata.jinja
```diff
@@ -67,38 +67,38 @@
                 Unknown
               {% endif %}
             {% else %}
-              —
+              No data available
```
anthonyhashemi: Could we utilise a template filter to centralise `No data available` across all of these?

jlhdxw: I have added that in now

anthonyhashemi (approved):
Nice one

PR #3088, opened 2026-06-22
Fclc 1828/update pagination styling

jlhdxw:
## Changes in this PR:

Updates the pagination to be in the same style as govuk.

https://national-archives.atlassian.net/browse/FCLC-1828


ds_caselaw_editor_ui/templates/components/pagination.jinja
```diff
@@ -1,73 +1,45 @@
...
-            <span>Next page</span>
-          </span>
-        {% endif %}
-      </li>
-    </ul>
-  </nav>
+{% from "govuk_frontend_jinja/components/pagination/macro.html" import govukPagination %}
+{% macro pagination(params) %}
+  {{ govukPagination(params) }}
+{% endmacro %}
```
anthonyhashemi, line 4: [non-blocking q]: Do we need to maintain a copy of the govuk pagination macro in our own components page if we just override the styling?

anthonyhashemi (approved):

ds_caselaw_editor_ui/templates/components/pagination.jinja
jlhdxw: That's the pattern we use for all these govuk macros so we have a single interface style for pulling components, so if another developer comes they know to just use components from the component list, rather than not knowing whether to use a govuk one or an internal one or something else.

PR #3206, opened 2026-07-16
FCLC-2116 | add validation for when a upload is attempted to a publis…

jlhdxw:
…hed document

## Changes in this PR:

Adds validation for when an editor tries to upload a document to a document that has been published.

https://national-archives.atlassian.net/browse/FCLC-2116


e2e_tests/test_judgment_page.py
```diff
@@ -39,6 +40,9 @@ def test_judgment_upload(authenticated_page: Page):
     authenticated_page.goto(f"{JUDGMENT_URI}/upload")
     assert_matches_snapshot(authenticated_page, "judgment_upload_page")
```
anthonyhashemi, line 42: Any benefit having these 2 different scenarios under the same test? I'd suggest separate tests for easier debugability/separation of concerns

jlhdxw: Good point - have split it out

anthonyhashemi (approved):

PR #3216, opened 2026-07-17
Display html for docs with no header in xml

anthonyhashemi:
https://national-archives.atlassian.net/browse/FCLC-2158

 by updating ds-caselaw-ma…rklogic-api-client to version 47.2.1


dynamictulip (approved):

jacksonj04:
This will be resolved by upcoming APIC version bumps, and doesn't need to be done separately.

PR #3293, opened 2026-08-05
[FCLC-2254] Add document metadata page with per-field claim provenance

jacksonj04:
This introduces a new "Metadata" page for each document, where the entire provenance chain for all the document's metadata can be inspected. This is intended to support the building of the editable metadata stack.

Navigation is behind the `document_metadata` flag for now.

## Jira

FCLC-2254


jlhdxw (approved):
Looks good as a first pass - we can adjust the page based off design chats later

PR #3328, opened 2026-08-10
Metadata framework rendering improvements

jacksonj04:
A series of improvements to how we render stuff using the updated metadata framework:

- Clearer sections in the metadata sidebar about where stuff comes from
- Page titles actually lean on the new metadata layer, not the old `Document` property
- We no longer display fallback body-based metadata values where we have materialised ones

### DIFF
--- a/ds_caselaw_editor_ui/sass/components/_aside.scss
+++ b/ds_caselaw_editor_ui/sass/components/_aside.scss
@@ -117,6 +117,26 @@
           margin: 0;
         }
       }
+
+      .aside__section-heading {
+        gap: 0;
+        margin-top: $space-2;
+        padding-top: $space-3;
+        border-top: 1px solid colour-var("accent-border");
+
+        &:first-child {
+          margin-top: 0;
+          padding-top: 0;
+          border-top: 0;
+        }
+
+        dt {
+          font-size: $typography-sm-text-size;
+          font-weight: $typography-bold-font-weight;
+          color: colour-var("font-base");
+          text-transform: uppercase;
+        }
+      }
     }
 
     @media (min-width: $grid-breakpoint-medium) {
--- a/judgments/tests/test_metadata_panel.py
+++ b/judgments/tests/test_metadata_panel.py
@@ -145,6 +145,54 @@ class TestMetadataPanel(TestCase):
 
         assert root.xpath('//dt[text()="TDR ref"]/following-sibling::dd')[0].text_content().strip() == "TDR-999"
 
+    def test_document_metadata_aside_shows_other_identifiers_without_preferred(self):
+        template = environment(loader=PackageLoader("ds_caselaw_editor_ui", "templates")).from_string(
+            '{% from "components/document_metadata.jinja" import document_metadata %}'
+            "{{ document_metadata(document=document) }}",
+        )
+
+        only = SimpleNamespace(
+            schema=SimpleNamespace(name="Find Case Law Identifier"),
+            value="tn4t35ts",
+            uuid="id-only",
+        )
+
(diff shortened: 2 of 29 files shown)


judgments/tests/test_metadata_panel.py
```diff
@@ -62,6 +65,140 @@ def test_metadata_item_renders_multi_value_badges(self):
...
+        template = environment(loader=PackageLoader("ds_caselaw_editor_ui", "templates")).from_string(
+            '{% from "components/document_metadata_item.jinja" import document_metadata_item %}{{ document_metadata_item(metadata_item=metadata_item) }}',
+        )
+
+        rendered = template.render(
+            metadata_item=SimpleNamespace(values=[DocumentCategory(name="Civil"), DocumentCategory(name="Tax")]),
+        )
+
+        assert '<div class="badge badge--info ">Civil</div>' in rendered
+        assert '<div class="badge badge--info ">Tax</div>' in rendered
```
jlhdxw: I don't think we should test the HTML is as expected - the issue is when we change the html, this test fails, where as what we care about is the "civil" and "tax" are rendered.

judgments/tests/test_metadata_panel.py
```diff
@@ -62,6 +65,140 @@ def test_metadata_item_renders_multi_value_badges(self):
...
+            has_ever_been_published=False,
+            first_published_datetime_display=None,
+            source_name="",
+            source_email="",
+        )
+
+        rendered = template.render(document=document)
+        root = lxml.html.fromstring(rendered)
+
+        section_headings = root.xpath('//div[contains(@class, "aside__section-heading")]/dt/text()')
```
jlhdxw: Same comment here WRT the class name. The class name isn't something we should be including in this kind of test as it's a display concern which should be able to change without this test failing. If we need something to hook onto for this page, we can add a test id (data-test-id="metadata-section-heading" or something)

judgments/tests/test_metadata_panel.py
```diff
@@ -62,6 +65,140 @@ def test_metadata_item_renders_multi_value_badges(self):
...
+        judgment.metadata["judges"] = SimpleNamespace(values=["Lord Justice Underhill"])
+        mock_judgment.return_value = judgment
+
+        self.client.force_login(User.objects.get_or_create(username="testuser")[0])
+
+        response = self.client.get(
+            reverse("full-text-html", kwargs={"document_uri": judgment.uri}),
+        )
+
+        self.assertContains(response, '<div class="aside__section-heading">')
```
jlhdxw: Same comment here WRT testing display things

ds_caselaw_editor_ui/sass/components/_aside.scss
```diff
@@ -117,6 +117,27 @@
...
+        border-top: 1px solid colour-var("accent-border");
+
+        &:first-child {
+          margin-top: 0;
+          padding-top: 0;
+          border-top: 0;
+        }
+
+        dt {
+          font-size: 0.875rem;
```
jlhdxw: We should have values for sizes like this (or if not, use whatever is closest)

ds_caselaw_editor_ui/sass/components/_aside.scss
```diff
@@ -117,6 +117,27 @@
...
+          font-weight: $typography-bold-font-weight;
+          color: colour-var("font-base");
+          text-transform: uppercase;
+          letter-spacing: 0.04em;
```
jlhdxw: Same with this - if we don't have a variable, we probably shouldn't do this.

jlhdxw (approved):
LGTM

